PR #34, opened 2015-02-10
User friendly error pages.

rjkip:
Don't need no SAML flow resetting in the Gateway, correct?


rjkip:
Assuming this is OK, but pls :+1:

DRvanR:
:+1:

PR #35, opened 2015-02-11
SSO Session implementation

rjkip:
We will not keep a (SSO) session on the gateway. Thus:
* Each SP authentication will trigger an authentication request at the IdP (e.g. SURFconext). The IdP may offer SSO.
* Each authentication > LoA 1 will trigger an authentication of the 2nd factor

We will not forward the SSO session information from the IdP to the SSP, or provide session information to the SP. So in the Response from GW to SP:
* No SessionNotOnOrAfter in the AuthnStatement
* No SessionIdentifier in the AuthnStatement

The assertion is to be valid for 5 minutes. 

See https://wiki.surfnet.nl/display/stepup/SAML+2.0+messages for sample SAML messages

---

Checked assertion expiration: https://github.com/SURFnet/Stepup-Gateway/blob/01df5457a82af8c6d8c13b4ee32914331d442b9e/src/Surfnet/StepupGateway/GatewayBundle/Service/ProxyResponseService.php#L94
Ensured LOA>1 = trigger 2nd factor: https://github.com/SURFnet/Stepup-Gateway/blob/01df5457a82af8c6d8c13b4ee32914331d442b9e/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php#L149

### DIFF
--- a/build-pre-commit.xml
+++ b/build-pre-commit.xml
@@ -60,13 +60,17 @@
         <loadresource property="changeset.php.absolute.notests">
             <propertyresource name="changeset.php.absolute.newlinesep"/>
             <filterchain>
-                <tokenfilter>
-                    <containsregex pattern="src/(?!Tests).*$"/>
-                </tokenfilter>
-                <deletecharacters chars="\n"/>
+                <linecontains negate="true">
+                    <contains value="Tests"/>
+                </linecontains>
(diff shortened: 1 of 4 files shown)


DRvanR:
awwww yissss

:+1:

PR #37, opened 2015-02-19
Break when SP requests invalid LoA

DRvanR:
Because the response was never returned it was possible to request a login with an invalid LoA.


DRvanR:
uhoh

@rjkip

rjkip:
:-1:... :+1: .

PR #45, opened 2015-03-24
Integrate SecondFactorType value object

rjkip:


rjkip:
:facepunch: @DRvanR

DRvanR:
:+1:

PR #47, opened 2015-03-30
Feature/translation fixes

pmeulen:
Fix textual issues in gateway


src/Surfnet/StepupGateway/GatewayBundle/Resources/views/SecondFactor/verifySmsSecondFactor.html.twig
```diff
@@ -9,7 +9,6 @@
 
     <ol>
         <li>{{ 'gateway.second_factor.sms.text.ensure_phone_has_signal'|trans }}</li>
-        <li>{{ 'gateway.second_factor.sms.text.enter_phone_number_below'|trans }}</li>
```
DRvanR: We need to ensure this is also removed from the xliff files.

pmeulen: Manually removed the now unused "gateway.second_factor.sms.text.enter_phone_number_below" from the xliff

DRvanR:
:+1:

PR #57, opened 2015-05-26
Take IdP LoA requirements into consideration

rjkip:
Take IdP LoA requirements into consideration when determining the which LoA to require the user to attain.

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
@@ -35,7 +35,7 @@ class ResponseContext
     private $hostedIdentityProvider;
 
     /**
-     * @var SamlEntityService
+     * @var \Surfnet\StepupGateway\GatewayBundle\Service\SamlEntityService
      */
     private $samlEntityService;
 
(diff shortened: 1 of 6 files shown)


src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
```diff
@@ -162,7 +167,9 @@ public function saveAssertion(SAML2_Assertion $assertion)
 
         // same for the entityId of the authenticating Authority
         $authenticatingAuthorities = $assertion->getAuthenticatingAuthority();
-        if (!empty($authenticatingAuthorities)) {
+        if (empty($authenticatingAuthorities)) {
+            $this->stateHandler->setAuthenticatingIdp($assertion->getIssuer());
+        } else {
             $this->stateHandler->setAuthenticatingIdp(reset($authenticatingAuthorities));
```
rjkip: I added this because SimpleSAMLphp doesn't include any authenticating authorities. Should we fall back to the message issuer or should IdPs support authenticating authorities? If they don't, they circumvent SP/IdP LoA requirements.

rjkip:
:sparkles: @DRvanR

src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
DRvanR: Afaik, we should not do this. If SSP is an issue (dev?) we should fix that :wink:

DRvanR:
Bar the comment, :+1:

rjkip:
Merged into develop.

PR #58, opened 2015-05-28
Display the Gateway in the locale of the second factor that is currently being verified

rjkip:


rjkip:
:bell: @DRvanR

rjkip:
And then? @DRvanR

DRvanR:
:+1:

PR #59, opened 2015-06-02
Move logging config to logging.yml and switch to syslog

pmeulen:
Removed monolog from config_prod.yml and put the complete logging related config in logging.yml
Added handlers for syslog and disabled graylog and file based options.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -1,6 +1,52 @@
 imports:
     - { resource: config.yml }
 
+# Be careful not to remove the prod-signaler handler, which overwrites
+# the prod-signaler handler defined in logging.yml. The handler defined
+# in logging.yml disables bubbling which means that none of the handlers
+# below are invoked. Since the current dev setup is incompatible with the
+# prod setup defined in logging.yml, this means we won't see any logs in
+# the infrastructure currently used (graylog). Overwriting the handler
+# here resolves that and reinstates the dev logging setup.
+#
+# this configuration must be replaced to reflect production setup
+# see https://www.pivotaltracker.com/story/show/96056010
+#
+monolog:
+    channels: [authentication]
+    handlers:
+        prod-signaler:
+            type: group
+            members:
+                - main_graylog
+                - main_logfile
+                - main_debuglog
+        main_graylog:
+            type:    buffer
+            handler: graylog
+            level:   NOTICE
+        main_logfile:
+            type:    stream
+            handler: logfile
+            level:   NOTICE
+            path:    %kernel.logs_dir%/%kernel.environment%.log
+        main_debuglog:
+            type:    stream
+            handler: logfile
+            level:   DEBUG
+            path:    "%kernel.logs_dir%/%kernel.environment%.debug.log"
+        graylog:
+            type:      gelf
+            publisher: { hostname: %graylog_hostname% }
+            formatter: surfnet_stepup.monolog.full_message_exception_gelf_message_formatter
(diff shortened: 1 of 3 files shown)


DRvanR:
Seems that it all can be simplified to:

```yml
monolog:
    channels: [authentication]
    handlers:
        prod-signaler:
            type: fingers_crossed
            action_level: ERROR
            passthru: NOTICE # this means that all message of level NOTICE or higher are always logged
            handler: main_syslog
            bubble: false # if we handle it, nothing else should
        main_syslog:
            type: syslog
            ident: stepup-gateway
            facility: user
            channels: ["!authentication"] # the auth channel is logged by the next handler
        authenthentication_syslog:
            type: syslog
            ident: stepup-gateway
            facility: user
            level: INFO
            channels: [authentication]
```
caveat: have not tested it :wink:

pmeulen:
Yes, it can be simplified, and I think your version is mostly the same. I kept the more complex setup as a form of documentation on how other logging configurations the be created

What is different are:

- the added `channels: ["!authentication"]`to the nested "main_syslog" handler. Is this indeed required?
- `passthru: NOTICE` in the prod-signaler

DRvanR:
There is another difference: the `group` handler (which just proxys to a single handler) and the `buffered` handler (which is no longer needed) are removed :wink:

The `channels: ["!authentication"]` ensures that no message from the authentication channel is logged by this handler, regardless of level. Currently they're not logged due to the level we're using on the authentication logging (`info`), but should that level change or the `action_level` on the `fingers_crossed` handler change, then they will be logged twice: once by the `main_syslog` handler and once by the `authentication_syslog` handler. This ensures that they can be logged separately. 
_TL;DR: no it is not required, yes it would be wise._

The `passthru` is used to identify which levels between the `passthru` and `action_level` settings (in this case `NOTICE` and `WARNING`) are always logged by the defined handler (`main_syslog`). Only when the `action_level` or higher is reached, the other levels will be logged. 

With respect to the commented configs, I just left that out for brevity :wink: It might be useful to link to the actual documentation available on the [symfony website][1] and the [monolog github repo][2].

With respect to the removed handlers, I removed those because a more complex setup makes it harder to read and determine what is going on, it obfuscates what the actual logging strategy is. I would advise to keep the configuration as simple as can be to ease maintenance and adoption.

[1]: http://symfony.com/doc/current/cookbook/logging/index.html
[2]: https://github.com/Seldaek/monolog

pmeulen:
"passthru" is an unrecognized option for fingers_crossed. What should this be (and where to find such info?)

DRvanR:
sorry, should be `passthru_level`. Most extensive configuration documentation is in the configuration comments sadly, see [here][1]

[1]: https://github.com/symfony/MonologBundle/blob/master/DependencyInjection/Configuration.php#L90

pmeulen:
- Simplified logging config per @DRvanR 's suggestion
- Moved `channels: ["!authentication"]` from nested "main_syslog" handler to prod-signaler, otherwise the main still handles the authentication channel.
- `passthru` -> `passthru_level`
- Tested with prod config

DRvanR:
:+1:

DRvanR:
Going to hold off merging till tomorrow morning, first need to test if it requires changes for dev env to keep working and if so implement those.

DRvanR:
@pmeulen I've changed the dev logging setup to be backwards compatible with regards to the current setup. Any objections/remarks? If not I'll merge this :smile:

PR #61, opened 2015-06-03
Always check the configured loa's to determine the required Loa

DRvanR:


DRvanR:
:8ball: @rjkip

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -93,13 +94,31 @@ public function __construct(
 
     /**
      * @param string          $identityNameId
-     * @param string          $requestedLoa
-     * @param ServiceProvider $serviceProvider
-     * @param IdentityProvider|null $authenticatingIdp
+     * @param string          $requiredLoa
```
rjkip: I believe this should be a `Loa`.

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -93,13 +94,31 @@ public function __construct(
...
+        $candidateSecondFactors = $this->secondFactorRepository->getAllMatchingFor($requiredLoa, $identityNameId);
+        $this->logger->info(
+            sprintf('Loaded %d matching candidate second factors', count($candidateSecondFactors))
+        );
+
+        return $candidateSecondFactors;
+    }
+
+    /**
+     * @param                  $requestedLoa
```
rjkip: Please add type `string`.

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -139,33 +158,10 @@ public function determineViableSecondFactors(
...
-
-    /**
-     * @param ArrayCollection $loaCandidates
-     * @return null|\Surfnet\StepupBundle\Value\Loa
-     */
-    private function resolveHighestLoa(ArrayCollection $loaCandidates)
-    {
         $actualLoas = new ArrayCollection();
         foreach ($loaCandidates as $loaDefinition) {
             $loa = $this->loaResolutionService->getLoa($loaDefinition);
```
rjkip: Doesn't this mean that when a SP requires a malformed LoA 'http://malformed.loa/loa5' this is ignored, and we fall back to a potentially lower LoA? Is that acceptable?

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -139,33 +158,10 @@ public function determineViableSecondFactors(
...
 
-        $highestLoa = $this->resolveHighestLoa($loaCandidates);
-        if (!$highestLoa) {
-            $this->logger->info(sprintf(
-                'Out of %d candidate LoA\'s, no highest LoA could be determined; no candidate second factors',
-                count($loaCandidates)
-            ));
-            return new ArrayCollection();
+        if (!count($loaCandidates)) {
+            throw new RuntimeException('No loa can be found, at least one Loa (SP default) should be found');
```
rjkip: I must say you're pushing my buttons writing LoA differently from log message to log message ('loa', 'LOA', 'Loa'). Can we stick to one spelling? :grin:

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
DRvanR: yes, then the loa will be determined based on configuration, however this should not occur. Since an unknown loa is requested, we cannot comply. Do note that in the request parsing (before handing over to surfconext for identification) we already [check if the requested loa exists][1]. 

[1]: https://github.com/SURFnet/Stepup-Gateway/blob/bugfix/no-longer-short-circuit-loa1/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php#L69

rjkip:
Pending CI, :+1:

PR #68, opened 2015-06-18
Verify numeric Yubikey public IDs

rjkip:
- [x] Switch StepupBundle back to `dev-develop`

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
@@ -232,7 +232,7 @@ class StepUpAuthenticationService
         $otp = YubikeyOtp::fromString($command->otp);
         $publicId = YubikeyPublicId::fromOtp($otp);
 
-        if ($publicId->getYubikeyPublicId() !== $secondFactor->secondFactorIdentifier) {
+        if (!$publicId->equals(new YubikeyPublicId($secondFactor->secondFactorIdentifier))) {
             return new YubikeyOtpVerificationResult(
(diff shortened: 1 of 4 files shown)


rjkip:
:mega: @DRvanR

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -228,16 +229,16 @@ public function verifyYubikeyOtp(VerifyYubikeyOtpCommand $command)
             return new YubikeyOtpVerificationResult(YubikeyOtpVerificationResult::RESULT_OTP_VERIFICATION_FAILED, null);
         }
 
-        $otp = Otp::fromString($command->otp);
+        $publicId = YubikeyPublicId::fromModHex(Otp::fromString($command->otp)->publicId)->getYubikeyPublicId();
```
DRvanR: Might want to split over several lines for readability :wink:
"the result of a methodcall on the result of a factorymethod taking the value of a public property of another object is used as input of a factorymethod whose result exposes a method used to get the actual value" is a lot for a oneliner hehe

imo something like the following makes it clearer

```php
$otp = Otp::fromString($command->otp);
$yubikeyPublicId = YubikeyPublicId::fromModHex($otp->publicId);

$publicId = $yubikeyPublicId->getYubikeyPublicId()
``` 

Alternatively, create an additional layer (wait wut?) that acts as a translator (`OtpToPublicIdConverter::stringToYubikeyPublicId($command->otp)`, both as an anticorruption layer (integration point) and to be able to better document as to why you *must* use the `Otp` object first, rather than injecting the string from the command directly (validation of full Otp format, extraction of publicId part). Just a thought.

DRvanR:
:+1: pending CIs

PR #78, opened 2015-09-02
U2F Server Api Implementation

rjkip:
* [x] Revert Stepup dependency feature branches to `dev-develop`
 * [x] fixup commits


DRvanR:
please don't call branches "u2f-monster".

DRvanR:
is this WiP or ready for review?

rjkip:
@DRvanR This is now ready for review

src/Surfnet/StepupGateway/ApiBundle/Controller/U2fVerificationController.php
```diff
@@ -0,0 +1,106 @@
...
+            return new JsonResponse(['errors' => [$e->getMessage()]], Response::HTTP_INTERNAL_SERVER_ERROR);
+        }
+
+        return new JsonResponse(['status' => 'SUCCESS'], Response::HTTP_OK);
+    }
+
+    /**
+     * @return U2fVerificationService
+     */
+    private function getService()
```
DRvanR: which service are you getting?

src/Surfnet/StepupGateway/ApiBundle/Controller/U2fVerificationController.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public function registerAction(
+        U2fRegisterRequest $registerRequest,
+        U2fRegisterResponse $registerResponse,
+        Requester $requester
+    ) {
+        try {
+            $result = $this->getService()->verifyRegistration($registerRequest, $registerResponse, $requester);
+        } catch (Exception $e) {
+            return new JsonResponse(['errors' => [$e->getMessage()]], Response::HTTP_INTERNAL_SERVER_ERROR);
```
DRvanR: Seem to be missing validation of the request. If I make a properly structured request with all empty strings as values for the scalars I get an internal server error rather than a 400. Or am I missing something?

src/Surfnet/StepupGateway/ApiBundle/Controller/U2fVerificationController.php
```diff
@@ -0,0 +1,106 @@
...
+            return new JsonResponse($result, Response::HTTP_OK);
+        }
+
+        return new JsonResponse($result, Response::HTTP_BAD_REQUEST);
+    }
+
+    public function revokeRegistrationAction(RevokeRequest $revokeRequest, Requester $requester)
+    {
+        try {
+            $this->getService()->revokeRegistration($revokeRequest, $requester);
```
DRvanR: what if I request revocation of a keyhandle that does not exist? Should also be a 400 no?

src/Surfnet/StepupGateway/ApiBundle/Service/U2fVerificationService.php
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /**
+     * @param U2fRegisterResponse $response
+     * @return RegisterResponse
+     */
+    private function adaptRegisterResponse(U2fRegisterResponse $response)
+    {
+        $adapted = new RegisterResponse();
+        $adapted->errorCode = $response->errorCode ?: 0;
```
DRvanR: what is this `0`? (please don't use magic numbers :wink: )

src/Surfnet/StepupGateway/ApiBundle/Service/U2fVerificationService.php
```diff
@@ -0,0 +1,256 @@
...
+                'An exception was thrown while verifying the U2F device authentication (%s: %s)',
+                get_class($e),
+                $e->getMessage()
+            );
+            $this->logger->critical($errorMessage, ['exception' => $e]);
+
+            throw new RuntimeException($errorMessage, 0, $e);
+        }
+
+        $apiResult = new U2fAuthenticationVerificationResult();
```
DRvanR: seems this code (and it's siblings in other places :wink:) can be made quite a lot simpler with something like:

```php
<?php

// includes mapping from $result status to internal status, unknown result leads to LogicException
$apiResult = U2fAuthenticationVerificationResult::fromVerificationResult($result);

if (!$result->wasSuccussful()) {
    $this->logger->error(sprintf('U2F authentication verification failed, reason: "%s"', $apiResult->getStatus()));
} else {
    $this->logger->notice('U2F authentication verification successful');
}

return $apiResult
```

src/Surfnet/StepupGateway/ApiBundle/Service/U2fVerificationService.php
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        $this->logger->notice('Revoked U2F device registration');
+    }
+
+    /**
+     * @param U2fRegisterRequest $request
+     * @return RegisterRequest
+     */
+    private function adaptRegisterRequest(U2fRegisterRequest $request)
```
DRvanR: seeing all the conversions everywhere, might consider making an adapter object which then can also be tested for correctness.

src/Surfnet/StepupGateway/U2fVerificationBundle/Controller/U2fController.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+final class U2fController extends Controller
+{
+    public function appIdAction()
+    {
+        return new JsonResponse(
```
DRvanR: please add a link to the docs that explain this format. Might even just add a json file to the webroot t.b.h., faster than a controller invocation :wink:

app/DoctrineMigrations/Version20150902154411.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('CREATE TABLE registration (key_handle VARCHAR(255) NOT NULL, public_key VARCHAR(255) NOT NULL, sign_counter INT NOT NULL, PRIMARY KEY(key_handle)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
```
DRvanR: For all: please add schema-name, prevents this to be ran on the wrong schema and at the same time guarantees that the executor has correct rights. Keep in mind that the schema is configurable. 

Also, not knowing the spec, is 255 chars enough for a public key? Might want to up that a little. The keyhandle seems to be a sha256, so that might actually be a bit smaller? Not sure if the latter is worth the hassle, but the former might be an issue...

src/Surfnet/StepupGateway/U2fVerificationBundle/Service/RegistrationVerificationResult.php
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * @SuppressWarnings(PHPMD.TooManyMethods)
+ */
+final class RegistrationVerificationResult
+{
+    /**
+     * @var U2fRegistrationVerificationResult
+     */
+    private $wrappedResult;
```
DRvanR: if it looks like a duck, acts like a ducks and sounds like a duck then call it a duck...

src/Surfnet/StepupGateway/U2fVerificationBundle/Service/VerificationService.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @param RegisterRequest  $request The register request that you requested earlier and was used to query the U2F
+     *     device.
+     * @param RegisterResponse $response The response that the U2F device gave in response to the register request.
+     * @return RegistrationVerificationResult
+     */
+    public function verifyRegistration(RegisterRequest $request, RegisterResponse $response)
+    {
+        $result = RegistrationVerificationResult::wrap($this->u2fService->verifyRegistration($request, $response));
```
DRvanR: honestly, these `RegistrationVerificationResult` and `AuthenticationVerificationResult` objects really seem like a step too far. They don't seem to add anything bar a layer of indirection.

src/Surfnet/StepupGateway/U2fVerificationBundle/Repository/RegistrationRepository.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param Registration $registration
+     */
+    public function save(Registration $registration)
+    {
+        $entityManager = $this->getEntityManager();
+        $entityManager->persist($registration);
+        $entityManager->flush();
+    }
+
+    public function revokeByKeyHandle(KeyHandle $keyHandle)
```
DRvanR: `removeByKeyHandle`? A repository has no conceptual knowledge of revocation :wink:

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Service/VerificationServiceAuthenticationVerificationTest.php
```diff
@@ -0,0 +1,123 @@
...
+            ->once()
+            ->with(m::anyOf(new KeyHandle($keyHandle)))
+            ->andReturn($registration);
+        $registrationRepository->shouldReceive('save')->never();
+
+        $u2fService = m::mock('Surfnet\StepupU2fBundle\Service\U2fService');
+        $u2fService
+            ->shouldReceive('verifyAuthentication')
+            ->with(m::anyOf($registrationDto), m::anyOf($request), m::anyOf($response))
+            ->once()
```
DRvanR: I really wouldn't test for whether or not the u2fService has this method invoked with any parameters. What is being tested is that if this service returns a failed response a method in another class is not invoked. That is the only expectation that should be present, otherwise there is more than 1 unit under test. 

Allmost rieks as if Behat tests would be more appropriate: given a certain request and a certain DB state, a certain response is and certain DB state are expected. Much bigger than just testing a single unit, you're testing the boundary interactions it seems. Thoughts?

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Service/VerificationServiceRegistrationRevocationTest.php
```diff
@@ -0,0 +1,47 @@
...
+use Surfnet\StepupGateway\U2fVerificationBundle\Value\PublicKey;
+use u2flib_server\Error;
+
+final class VerificationServiceRegistrationRevocationTest extends TestCase
+{
+    /**
+     * @test
+     * @group registration-revocation
+     */
+    public function it_can_revoke_a_registration_by_key_handle()
```
DRvanR: Don't think this test really adds anything. It tests internal communication between objects, while interacting outside of that communication. This indeed seem to be more behavioral tests (given I try this, the result should be this answer and that state) than unittests.

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Value/ValueObjectTest.php
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupGateway\U2fVerificationBundle\Tests\Value;
+
+trait ValueObjectTest
```
DRvanR: it's a dataprovider, not a test.

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Value/KeyHandleTest.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+namespace Surfnet\StepupGateway\U2fVerificationBundle\Tests\Value;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Surfnet\StepupGateway\U2fVerificationBundle\Value\KeyHandle;
+
+final class KeyHandleTest extends TestCase
+{
+    use ValueObjectTest;
```
DRvanR: Might be worth to introduce this in the Middleware as well as we've duplicated these providers there all over the place.

src/Surfnet/StepupGateway/U2fVerificationBundle/Value/PublicKey.php
```diff
@@ -0,0 +1,58 @@
...
+     * @var string
+     */
+    private $publicKey;
+
+    /**
+     * @param string $publicKey
+     */
+    public function __construct($publicKey)
+    {
+        if (!is_string($publicKey)) {
```
DRvanR: public keys may be empty strings?

src/Surfnet/StepupGateway/U2fVerificationBundle/Value/KeyHandle.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @param string $keyHandle
+     */
+    public function __construct($keyHandle)
+    {
+        if (!is_string($keyHandle)) {
+            throw InvalidArgumentException::invalidType('string', 'keyHandle', $keyHandle);
+        }
+
+        if (empty($keyHandle)) {
```
DRvanR: please use `trim($string) == ''` to test for empty strings, `'0'` is considered empty while it isn't.

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Value/ValueObjectTest.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function nonStrings()
+    {
+        return [
+            'int'      => [1],
+            'float'    => [1.1],
+            'resource' => [fopen('php://memory', 'w')],
+            'object'   => [new \stdClass],
+            'array'    => [array()],
+            'bool'     => [false],
```
DRvanR: we might want to add null bytes to these (`chr(0)`) to see how well we've checked :smile_cat:

app/DoctrineMigrations/Version20150902154411.php
rjkip: I [dove](https://fidoalliance.org/specs/fido-u2f-v1.0-nfc-bt-amendment-20150514/fido-u2f-raw-message-formats.html#registration-response-message-success) in the spec, and the public key is always a 65-byte string, Base64-encoded, resulting in a string that is never more than 88 characters long. I cannot find any requirements of the key handle length other than that, in the registration response, the key handle length is specified using one unsigned byte, ie. 0-255.

Thus, I could resize the public key column to a maximum of 88 characters. Is there value in doing this other than being correct?

src/Surfnet/StepupGateway/U2fVerificationBundle/Tests/Value/ValueObjectTest.php
rjkip: I could test `KeyHandle` and `PublicKey` with null characters, but there's nothing in the spec about this and a string with a null character in it is perfectly valid IMO. If it isn't there surely are more characters than null-characters we should typically check for.

src/Surfnet/StepupGateway/U2fVerificationBundle/Value/KeyHandle.php
rjkip: Valid, will change.

src/Surfnet/StepupGateway/ApiBundle/Controller/U2fVerificationController.php
rjkip: Exceptions are not expected to occur and will result in a HTTP 500. Expected errors are handled through the result object and result in a HTTP 400. Further validation of device data will be addressed by another story.

rjkip:
@DRvanR Please review.

src/Surfnet/StepupGateway/U2fVerificationBundle/Service/VerificationService.php
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param KeyHandle $keyHandle
+     * @return bool Whether the registration was found and removed.
+     */
+    public function revokeRegistration(KeyHandle $keyHandle)
```
DRvanR: Reconsider it a bit please, as discussed.

rjkip:
That all? @DRvanR

DRvanR:
:+1:

PR #79, opened 2015-09-14
Enable entering of Yubikey OTPs in Internet Explorer browsers

rjkip:
This is achieved by placing the Yubikey OTP submit button off-screen
rather than by not displaying it (display:none). The latter solution
makes Internet Explorer browsers pick the cancel button as the first
submit button, canceling the authentication process.

Fixes [Yubikey authentication in IE11 activates the cancel button](https://www.pivotaltracker.com/story/show/103173876).

### DIFF
--- a/app/Resources/public/less/style.less
+++ b/app/Resources/public/less/style.less
@@ -38,6 +38,7 @@ form[name="gateway_verify_sms_challenge"] {
     }
 }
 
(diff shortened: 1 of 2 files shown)


rjkip:
@DRvanR Your care and diligence is required!

DRvanR:
:+1:

PR #81, opened 2015-09-23
Add ability to create U2F register and sign requests

rjkip:


rjkip:
:bomb: @DRvanR

DRvanR:
please change 201 -> 200, other than that :+1:

PR #83, opened 2015-09-24
Allow configuring of enabled second factors

rjkip:
image

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
@@ -77,13 +77,28 @@ class StepUpAuthenticationService
      */
     private $logger;
 
+    /**
+     * @var string[]
+     */
+    private $enabledSecondFactorTypes;
+
+    /**
+     * @param LoaResolutionService   $loaResolutionService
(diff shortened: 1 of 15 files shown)


rjkip:
kejfke

@DRvanR 

I'm off.

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -107,7 +123,23 @@ public function determineViableSecondFactors(
             sprintf('Loaded %d matching candidate second factors', count($candidateSecondFactors))
         );
 
-        return $candidateSecondFactors;
+        $enabledCandidateSecondFactors = new ArrayCollection();
+        foreach ($candidateSecondFactors as $candidate) {
```
DRvanR: Why get secondfactors from a Repo that are not allowed to be gotten only to filter them out afterwards? In my mind it makes more sense to let the SFRepo be aware of what is/isn't enabled, so that the repository can actively ensure that not-enabled secondfactors cannot be used anywhere, rather than having to enforce that afterwards. Best might even be a decorator around the current repo, to allow for filtering and the extended logging as done below. Just a thought.

rjkip:
d

@DRvanR 

d

DRvanR:
Very nice, :+1:

PR #85, opened 2015-09-29
Don't attempt to fetch second factor when id is null

rjkip:


rjkip:
@DRvanR 
dfd

DRvanR:
:1234: pick one.

DRvanR:
:+1:

PR #97, opened 2016-06-06
Stepup locale cookie

relaxnow:
@rjkip review?


app/config/config.yml
```diff
@@ -173,6 +173,13 @@ surfnet_stepup:
         originator: %sms_originator%
         otp_expiry_interval: %sms_otp_expiry_interval%
         maximum_otp_requests: %sms_maximum_otp_requests%
+    locale_cookie:
+        name: stepup_locale
+        domain: %locale_cookie_domain%
+        expire: 0
+        path: /
+        secure: true
+        httpOnly: false
```
rjkip: Any reason to not enable this?

src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
```diff
@@ -261,6 +261,24 @@ public function getResponseContextServiceId()
...
+     * @return $this
+     */
+    public function setPreferredLocale($locale)
+    {
+        $this->set('locale', $locale);
+        return $this;
+    }
+
+    /**
+     * @return string
```
rjkip: This can be null.

rjkip:
Nice and simple integration :+1:

app/config/config.yml
relaxnow: There are no secrets in it, seems useful to let client side code access the value.

rjkip: But then I must ask: which client-side code currently needs this? :no_mouth:

relaxnow:
@rjkip rereview?

rjkip:
:ship:

rjkip:
Pending build status, :+1::ship:

PR #98, opened 2016-06-07
Second Factor Only issues

relaxnow:
* Require Gateway AuthnRequests to be signed.
* Return NULL if Service Provider is not found.	
* Turn SFO nameid validation INFOs into NOTICEs

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
@@ -44,7 +44,7 @@ class GatewayController extends Controller
         $redirectBinding = $this->get('surfnet_saml.http.redirect_binding');
 
         try {
-            $originalRequest = $redirectBinding->processUnsignedRequest($httpRequest);
+            $originalRequest = $redirectBinding->processSignedRequest($httpRequest);
         } catch (Exception $e) {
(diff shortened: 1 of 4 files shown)


src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
```diff
@@ -44,7 +44,7 @@ public function ssoAction(Request $httpRequest)
         $redirectBinding = $this->get('surfnet_saml.http.redirect_binding');
 
         try {
-            $originalRequest = $redirectBinding->processUnsignedRequest($httpRequest);
+            $originalRequest = $redirectBinding->processSignedRequest($httpRequest);
```
rjkip: Hmm, I incorrectly assumed the change to unsigned requests was on purpose. :flushed:

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/SecondFactorOnlyNameIdValidationService.php
```diff
@@ -43,7 +43,7 @@ public function with(LoggerInterface $logger)
     public function validate($spEntityId, $nameId)
     {
         if (!$nameId) {
-            $this->logger->info(
+            $this->logger->notice(
```
rjkip: :+1:

rjkip:
:ship:

PR #100, opened 2016-06-08
Require PHP 5.6 and ran composer update

relaxnow:
Note this PR goes to stepup-locale-cookie because the stepup-bundle now requires locale_cookie settings with that PR already has.


composer.json
```diff
@@ -9,7 +9,7 @@
     },
     "minimum-stability": "stable",
     "require": {
-        "php": ">=5.4,<8.0-dev",
+        "php": "~5.6.0",
```
rjkip: I think you accidentally excluded 7.x here :smirk:

composer.json
```diff
@@ -26,14 +26,14 @@
         "nelmio/security-bundle": "~1.4",
         "doctrine/doctrine-migrations-bundle": "~1.0.0",
         "doctrine/migrations": "~1.0.0|1.0.0@alpha",
-        "mopa/bootstrap-bundle": "3.0.0-RC2",
+        "mopa/bootstrap-bundle": "3.0.0",
```
rjkip: Did you look into what changed between RC2 and stable and whether it affects us? Mopa Bootstrap has made many BC breaks during its journey to a stable 3.0.

composer.json
relaxnow: No, quite on purpose actually. SURFnet doesn't run Step Up on 7.0. So it may work, it may not work we don't know, if you want it to work you use 5.6 😄

rjkip: In that case, you ought to update the Travis build file to upgrade the PHP requirement to 7.x while testing. Otherwise the dependencies won't install.

composer.json
relaxnow: I didn't, it appears they have quite a few changes: https://github.com/relaxnow/MopaBootstrapBundle/pull/1

But I'm not well versed enough in the innards to decide whether this is 'safe'.

What do you think @rjkip ?

rjkip: I really have no idea. This requires going through the entire application and checking whether everything works in modern browsers. Something that requires special attention and not something to do while upgrading other things. Let's revert to RC2?

relaxnow: It looks like an upgrade to font-awesome 4.4 and compatibility with SF 3 is the bulk of it though.
Sounds like just something we want as we're trying to make SSA SF 3 compatible, right?

rjkip: Let's say I feel it's risky. I'd only upgrade after manual checking. Although it sounds like a good idea, I have no knowledge of a SURF-initiated push towards Symfony 3. The next LTS is released in November 2017.

composer.json
relaxnow: Good point! Forgot about ~~Dre~~ that! Fixed.

composer.json
relaxnow: It's part of https://www.pivotaltracker.com/projects/1163646/stories/109795400 to make SSA SF 3 compatible. But seeing as I need to test for that story I'll pick this up then and revert it for now.

relaxnow:
@rjkip :shipit: ?

composer.json
rjkip: Hmhmh kinda iffy, with `ignore-platform-reqs` we could potentially install PHP 5-only dependencies. With all these considerations we have to make, I'm leaning more and more to just going back to requiring PHP ~5.6|~7.0 so we can reliably test PHP 7 as well, or dropping testing of PHP 7 entirely. Also, shouldn't the tests prove (to a certain degree) that the application works?

relaxnow: If the tests pass for PHP 7.0 I still wouldn't want to recommend running it in production without significant manual testing.
So let's drop PHP 7 testing for now.

relaxnow: Out of band discussion: let's keep testing on PHP 7 but note in the README we don't officially support it for now.

rjkip:
:ship: :ship: :ship:

PR #101, opened 2016-06-16
FIx gssf authentication route assembly

relaxnow:
This bug was introduced in GW 1.4.0 and broke Tiqr and Biometric authentication.

GW 1.4.0 won't be deployed, so no need to backport it.

@rjkip could you review?

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Entity/SecondFactor.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Entity/SecondFactor.php
@@ -107,7 +107,7 @@ class SecondFactor
     /**
      * @return bool
      */
-    public function requiresGssf()
+    public function isGssf()
     {
         return (new SecondFactorType($this->secondFactorType))->isGssf();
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupGateway/GatewayBundle/Entity/SecondFactor.php
```diff
@@ -103,4 +103,12 @@ public function getLoaLevel()
     {
         return (new SecondFactorType($this->secondFactorType))->getLevel();
     }
+
+    /**
+     * @return bool
+     */
+    public function requiresGssf()
```
rjkip: I think a second factor *is* a Generic Saml Second Factor, and does not *require* it.

rjkip:
Other than a small thing, if you've tested the solution, :ship:.

relaxnow:
As a foolish programmer once said:
> If you haven't tested it, it probably doesn't work.

So I tested it and (now) it works 😈

PR #104, opened 2016-10-13
Remove locale cookie

arothuis:
[113611545](https://www.pivotaltracker.com/story/show/113611545)

This removes the locale cookie listener from the StepupBundle in Gateway.

A preferred locale cannot yet be determined when a second factor is being selected in Gateway, as at that point an intrinsic LoA is used, while setting the preferred locale is determined based on a (known) viable second factor. So Gateway determines it locale solely based on the identity's SecondFactor. It does not talk to MW by design. The displayed language is currently determined by the locale of SelfService and RA, which are stored in their sessions. MW/GW don't know about these locales upon entry. Furthermore, GW does not store any state about the identity. This is why the LocaleCookieListener will be implemented in [RA](https://github.com/SURFnet/Stepup-RA/pull/123) and in [SelfService](https://github.com/SURFnet/Stepup-SelfService/pull/111).

Related PRs:
- [Stepup-SelfService #111](https://github.com/SURFnet/Stepup-SelfService/pull/111)
- [Stepup-RA #123](https://github.com/SURFnet/Stepup-RA/pull/123)

This undos most of the work from #97.


arothuis:
Closing this as we need to use the SecondFactor's preferred locale when it has already been registered and we are not approaching the application through SS or RA.

PR #105, opened 2016-10-18
Do not sign / encrypt cookies by default

arothuis:
The locale cookie should be able to convey which locale has been chosen, obscuring it through the NelmioSecurityBundle would make this impossible

Whenever cookies need to be signed, one could do so explicitly, see: https://github.com/nelmio/NelmioSecurityBundle#signed-cookies


rjkip:
:ship:

PR #107, opened 2017-01-12
Bugfix/validate signatures indifferent to parameter encoding

arothuis:
SURFnet/Stepup-saml-bundle#55
https://www.pivotaltracker.com/story/show/136419577


arothuis:
As discussed, once the bugfix has been reviewed more extensively in the `stepup-saml-bundle`, we should upgrade that release and depend on that. After that, we should create a stable release for Gateway.

PR #122, opened 2017-09-07
Log LoA resolution messages as notices

jorissteyn:
Events logged in a happy-flow are logged with log level 'notice', so
events happening in an 'unhappy' flow should at least be a notice.


jorissteyn:
Merged as part of #121

PR #127, opened 2017-12-14
Institution specific LoA determination for unknown user

MKodde:
This is the bugfix for the problem described in [Pivotal](https://www.pivotaltracker.com/story/show/153594101)

The story describes the desired authentication flow when an SP/institution specific configuration applies. The possible situations are covered by a unit test of the `StepupAuthenticationServiceTest`.

@pmeulen can you think of other scenario's that should be covered by this unit test? The following tests are implemented at the moment:
* Happy flow: no sp-institution specific configuration provided
* Happy flow: an sp-institution specific configuration is provided, the correct loa is requested
* Missing schacHomeOrganization for authenticating user, sp-institution specific configuration is provided
* Sp-institution specific config is provided but the user does not possess a token that was vetted at that institution
* No default loa was configured for the SP
* User cannot provide viable token


src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
```diff
@@ -172,6 +172,15 @@ public function saveAssertion(SAML2_Assertion $assertion)
             $this->stateHandler->setAuthenticatingIdp(reset($authenticatingAuthorities));
         }
 
+        // And also attempt to save the user's schacHomeOrganization
+        $attributes = $assertion->getAttributes();
+        if (array_key_exists('urn:mace:terena.org:attribute-def:schacHomeOrganization', $attributes) &&
+            !empty($attributes['urn:mace:terena.org:attribute-def:schacHomeOrganization'])
```
jorissteyn: Simply checking for not empty is enough in this case

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -180,18 +182,48 @@ public function resolveHighestRequiredLoa(
...
+                'SP configured LOA\'s are applicable but the authenticating user has no ' .
+                'schacHomeOrganization in the assertion.'
+            );
+        }
+
+        // Load the institutions the user has vetted a second factor
+        $institutionsBasedOnVettedTokens = $this->determineInstitutionsByIdentityNameId($identityNameId);
+
+        if (!is_null($identityInstitution) && !empty($institutionsBasedOnVettedTokens)) {
+            $institutionMatches = $this->institutionMatchingHelper->findMatches(
```
jorissteyn: Perhaps move this to a private method, I found $institutionMatches in the same scope as $matchingInstitutions confusing.

pmeulen:
@MKodde 
Regarding tests.
1a. No sp-institution specific configuration provided (i.e. '__default__' = LoA 1)
1b. An sp-institution specific configuration is provided for institution <A> with a LoA = 2
2a. The user has a schacHomeOrganization attribute set
2b. The user has no schacHomeOrganization attribute set
2c. The user has a schacHomeOrganization attribute set that is different from the one used during registration
3a. The user has a vetted token (i.e. NameID exists in the second_factor table)
3b. The user does not not have a vetted token
4a. SP does not request a LoA (i.e. no AuthContexClassRef in AuthnRequest)
4b. SP requests LoA = 1
4c. SP requests LoA = 2
4c. SP requests LoA = 3

I suggest adding (well, generating, 2 * 3 * 2 * 4 = 48 tests) a test for each combination.

MKodde:
@pmeulen I'd like to discus one of described cases where

> SP institution specific config provided, user has SHO of existing institution but has got no registered tokens. (1b, 2a, 3b, 4a,b,c,d)

What sort of behavior do we expect here? I would expect the user would not be allowed to pass. Am I correct in my assumption?

pmeulen:
> @pmeulen I'd like to discus one of described cases where

> > SP institution specific config provided, user has SHO of existing institution but has got no registered tokens. (1b, 2a, 3b, 4a,b,c,d)

> What sort of behavior do we expect here? I would expect the user would not be allowed to pass. Am I correct in my assumption?

Yes, you are correct. This is precisely the case we are fixing in this issue. Because the LoA requirement (LoA = 2) is now found based on the SHO from the assertion, the user is denier access because the user cannot meet the LoA requirement.

src/Surfnet/StepupGateway/GatewayBundle/Service/StepUpAuthenticationService.php
```diff
@@ -350,8 +365,36 @@ public function clearSmsVerificationState()
...
+                    'None of the authenticating users tokens are registered at an institution the user is currently ' .
+                    'authenticating from.'
+                );
+            }
+        }
+        // Add the SHO of the authenticating user to the list of institutions based on vetted tokens. This ensures the
+        // correct LOA can be based on the organisation of the user.
+        // There is a chance the user has no vetted tokens. In that case the user will not be granted access to the SP
+        // if the LOA is higher than LOA1 @see https://www.pivotaltracker.com/story/show/153594101
+        $institutionsBasedOnVettedTokens[] = $identityInstitution;
```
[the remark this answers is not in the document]
MKodde: @pmeulen a question about the matrix. Column 10  (1a, 2a, c2, 4a, expected LoA: 2).

I personally don't see the logic in this outcome. The user

MKodde:
@pmeulen a question about the matrix. Column 10  (1b, 2a, 3b, 4a, expected LoA: 2).

I personally don't see the logic in this outcome.

Given this input into the method:
1b. An sp-institution specific configuration is provided for institution <A> with a LoA = 2
2a. The user has schacHomeOrganization attribute set to <A>
3b. The user does not not have a vetted token
4a. SP does not request a LoA (i.e. no AuthContexClassRef in AuthnRequest)

I would expect an exception as the outcome. And not LoA2. As the user belongs to the institution of which the SP has a custom LoA requirement. But the user does not have any tokens.

pmeulen:
> @pmeulen a question about the matrix. Column 10  (1b, 2a, 3b, 4a, expected LoA: 2).

> I personally don't see the logic in this outcome.

> Given this input into the method:
> 1b. An sp-institution specific configuration is provided for institution <A> with a LoA = 2
> 2a. The user has schacHomeOrganization attribute set to <A>
> 3b. The user does not not have a vetted token
> 4a. SP does not request a LoA (i.e. no AuthContexClassRef in AuthnRequest)

> I would expect an exception as the outcome. And not LoA2. As the user belongs to the institution of which the SP has a custom LoA requirement. But the user does not have any tokens.

The outcome is indeed an access denied, because the user cannot meet the required LoA of 2.

What I have put in the column is the "Resulting  Required LoA", in this case 2, the evaluation of whether the user can actually meet the LoA is not included in the result.

joostd:
I think there is a corner case left that results in incorrect responses: (1) default LoA 1, higher LoA for SHO <A>, (2) user has SHO _not_ matching <A>, (3) user does not have a vetted token, (4) user does not request a LoA.

Such users are now returned to the SP with a SAML response containing an IdP error (NoAuthnContext)

PR #130, opened 2018-01-16
Upgrade of the Stepup-saml-bundle

MKodde:
Installation of the Stepup-saml-bundle version ~~~3.0.0.rc2~~~ ^3.0. Followed by some code changes to prevent errors on backwards incompatible changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "jms/translation-bundle": "~1.1.0",
         "jms/di-extra-bundle": "~1.4.0",
         "mockery/mockery": "~0.9.0",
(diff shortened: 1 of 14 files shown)


jorissteyn:
I've found one place where we still use the PEAR-style SAML2 names:

```
Stepup-Gateway $ grep -ir 'SAML2_' src
src/Surfnet/StepupGateway/SamlStepupProviderBundle/Saml/ProxyResponseFactory.php:     * @return SAML2_Response
src/Surfnet/StepupGateway/SamlStepupProviderBundle/Saml/ProxyResponseFactory.php:        $response = new SAML2_Response();
```

PR #137, opened 2018-03-09
Don't clear state on SAML responses or error pages

jorissteyn:
When Stepup renders a SAML response or error page, the state of the
request is never reset. If for whatever reason the request to the
consume-assertion request happens twice, both requests succeed. Also,
if the user refreshes the error page, the exact same error
occurs. This is intended behaviour and a good thing.

When Stepup acts as a proxy for GSSP, the behaviour is different. The
state is cleared after consume-assertion and when rendering an error
page. When the error page is refreshed by the user, different errors
are triggered because the request does not match the state. For
example, this error might be logged when refreshing an error page:

    Invalid Argument, parameter "requestId" should be of type "string", "NULL" given

This commit changes the proxy controllers to never reset the state
when a request is handled. The behaviour in Stepup is now consistent.

### DIFF
--- a/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
+++ b/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
@@ -295,9 +295,6 @@ class SamlProxyController extends Controller
             'relayState' => $stateHandler->getRelayState()
         ]);
 
-        // clear the state so we can call again :)
-        $stateHandler->clear();
-
         return $response;
     }
 


src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
```diff
@@ -295,9 +295,6 @@ public function renderSamlResponse($view, StateHandler $stateHandler, SAMLRespon
             'relayState' => $stateHandler->getRelayState()
         ]);
 
-        // clear the state so we can call again :)
-        $stateHandler->clear();
```
pmeulen: Part of the state is the ID from the SAML AuthnRequest. This is used for replay protection. The ID from the AuthnRequest must match the InResponseTo in the Response. It is correct to invalidate the ID after receiving a SAML Response. So resending a SAML Response could result in a different error.

I agree that from a user and support perspective it is nice if reloading an error page does not change the error. But this behaviour can be achieved by other means (e.g. by redirecting to an error URL.

jorissteyn: Thanks, the proper solution is a but more complicated. I'm closing (and not merging) this PR and created a pivotal issue for further discussion on how to handle this in the future: https://www.pivotaltracker.com/story/show/155850805

PR #140, opened 2018-03-12
Remove SURFisms

MKodde:
- Updated the app name
- Removed SURFisms
- Replace the old logo with the new SURF secure id one


MKodde:
No problem, I'll fix it in a second.

PR #144, opened 2018-03-27
Authentication screens restyle

MKodde:
In these five commits the general OpenConext styling as seen in Engineblock is applied in StepUp. More importantly the authentication views have been revised to be more compact and contain less instructions for the end users. The design changes are based on the changes found on the [Surfnet wiki](https://wiki.surfnet.nl/display/coininfra/Nieuwe+schermen).

https://www.pivotaltracker.com/story/show/155856648

### DIFF
--- a/app/Resources/public/less/authentication.less
+++ b/app/Resources/public/less/authentication.less
@@ -10,11 +10,11 @@ body.authentication {
   }
 
   .container.authentication {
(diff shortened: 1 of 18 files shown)


jorissteyn:
Display on small screens was suboptimal, I took a stab at improving that.

PR #147, opened 2018-03-28
Allow unknown attributes to pass through Gateway

MKodde:
By upgrading the Steupup-saml-bundle to a version >2.10 we can enable the `ignore_unknown_attributes` feature for the gateway. Effectively fixing the bug 'described' in: 

https://www.pivotaltracker.com/story/show/156264642


pmeulen:
Verified working, and solving the issue

PR #148, opened 2018-03-29
Restyle error views

MKodde:
The error views are now presented in the new OpenConext style. The solution was to update the default styling to this design.

This PR was created on top of the #144 feature.

https://www.pivotaltracker.com/story/show/156343095

### DIFF
--- a/app/Resources/public/less/authentication.less
+++ b/app/Resources/public/less/authentication.less
@@ -1,24 +1,9 @@
-
 body.authentication {
-  background-color: @gray-lighter;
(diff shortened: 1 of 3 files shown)


jorissteyn:
Looks good, but needs rebasing on #144 to support small screen sizes.

And the horizontal line is missing in the footer:

screenshot_20180330_131708

Shouldn't we port this to SS and RA?

MKodde:
The HR was added in the Steupup-bundle. You might not have updated this bundle during review.

PR #164, opened 2018-08-08
Set Symfony environment from APP_DEV superglobal

MKodde:
The concept is to load the .env file from the application root. If it cannot be found there, try to fall back on a Stepup globally set .env. If this does not exist either, just commit the default values to PHPs environment.

The previously installed vlucas DotEnv component is used to load and parse existing .env files. In a perfect world we would utilize Symfony DotEnv. For now we're stuck on version 2.7 and have to make due with this component.

To test this, copy the contents of the `app_dev.php.dist` file to your `web/app_dev.php` file.

As to why this change is proposed. For functional testing, manipulating the environment Symfony runs in is required to put in place some test doubles for services used in the application. EngineBlock solves this by running ant tasks before and after running functional tests manipulating the 'env' value in the `app_dev.php` file by running a `sed` command. This effort could be the inspiration for updating EB in the future in that area.

@pablothedude, please give your opinion on this solution, as yesterday IRL we discussed a more high level solution that might be interesting as well!

:exclamation: This PR now also carries: #165 Provide a mock Yubikey service

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,17 @@ Run `app/console mopa:bootstrap:symlink:less` to configure Bootstrap symlinks.
 
 The Gateway is configured to only accept connections over SSL. Disable this under `nelmio_security` in `config.yml` or run the web server using a (self-signed) certificate.
 
+## Developer options
+
+### Mock Yubikey service
+If you are not in possession of an actual Yubikey device, using the Mock Yubikey service might prove useful. This
+mock service was created for end to end test purposes, but could be utilized in this situation. To use the mock service:
+
+1. Update your `src/Surfnet/StepupGateway/ApiBundle/Resources/config/services.yml`
+2. Find the `surfnet_gateway_api.service.yubikey` service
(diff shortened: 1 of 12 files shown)


pablothedude:
I think that it should be better to separate the logic set the ENV var and the logic to get the ENV var. I've run a couple of tests to check if it's possible to change the environment variable by cookie and that's possible which allows us to prevent getting state into the application.

You could test it yourself by putting the code below in the nginx gateway pool config:
```
        set $testenvok dev;
        if ($http_cookie ~* "testcookie" ) {
            set $testenvok test;
        }
        fastcgi_param APP_DEV $testenvok;
```

And putting the code below in the gateway php-fpm config:
`env[APP_DEV] = $APP_DEV`

This way we could set a cookie `testcookie` to allow us to set the environment in test mode without using state.

MKodde:
I've taken your proposal to heart and fixed this in OpenConext-deploy and this PR.

See: https://github.com/OpenConext/Stepup-Deploy/pull/72

PR #167, opened 2018-08-27
Symfony 3.4 upgrade

MKodde:
Todo:
 - [x] Replace dev- dependencies for surfnet-*-bundle dependencies.

Closing this PR, the actual merge of these changes will take place in #168

### DIFF
--- a/web/app.php
+++ b/web/app.php
@@ -2,14 +2,16 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
-
-require_once __DIR__.'/../app/AppKernel.php';
-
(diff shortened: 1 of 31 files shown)


web/app.php
```diff
@@ -2,14 +2,16 @@
 
 use Symfony\Component\HttpFoundation\Request;
 
-$loader = require_once __DIR__.'/../app/bootstrap.php.cache';
+$loader = require __DIR__.'/../app/autoload.php';
```
pablothedude: Can we do this already (< php 7) ?

MKodde: Yes I think we can

See https://symfony.com/blog/new-in-symfony-3-3-deprecated-the-classloader-component

PR #175, opened 2019-01-10
Fix sms challenge form

pablothedude:
The sms-challenge form had a bug which resulted in an error because
the required form element could not be loaded. By using the FQDN this
should be fixed.


pablothedude:
https://www.pivotaltracker.com/story/show/161619825

MKodde:
A leftover from #168

PR #179, opened 2019-05-09
Improve AuthNFailed GW response handling

MKodde:
The GW proxy controller now also returns the subcode and error message
when one of the GSSP returned a failed response.

The exception message that was thrown while processing the SAMLResponse
is set as the message yielding something like:

"Could not process received Response, error: "Cannot process response,
preconditions not met: "Responder/AuthnFailed User canceled the
request""

This is far from perfect, preferably we'd simply set the original
"User canceled the request" message, this would entail quite some
rework to the gateway.

https://www.pivotaltracker.com/story/show/164657123

### DIFF
--- a/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
+++ b/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
@@ -141,7 +141,8 @@ class SamlProxyController extends Controller
         } catch (ResponseFailureException $e) {
             $response = $this->createResponseFailureResponse(
                 $provider,
-                $this->getDestination($provider->getStateHandler())
+                $this->getDestination($provider->getStateHandler()),
+                $e->getMessage()
             );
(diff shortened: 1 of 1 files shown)


MKodde:
@pablothedude the change you added seems to do the trick. Some concerns:

1. This only changes the GSSP authentication behavior. Granted, that's what is described in the story, but during my initial investigation, I identified the other endpoints might also need this change in order to return the correct issuer.
2. Test coverage of this feature would be awesome, maybe we can whip up a simple Behat test that covers this kind of behavior. We should be able to configure our dummy GSSP device to return an AuthNFailed response quite easily.

pablothedude:
@MKodde I've created a issue for your raised concerns. Would you be so kind to elaborate on them:
https://www.pivotaltracker.com/story/show/166409479

MKodde:
@pmeulen what do you say about my first concern? TL/DR = I think the other endpoints might handle the AuthNFailed response incorrectly. Is this indeed a problem that you know of, or something you get as feedback from users of the platform?

MKodde:
Thanks @pablothedude and @pmeulen, let's merge!

PR #181, opened 2019-05-28
Update the saml-bundle, adding WantAuthnRequestsSigned requirement to metadata

pablothedude:
Add WantAuthnRequestsSigned="true" to the IDPSSODescriptor in the
Metadata to support service providers that require that attribute.

https://www.pivotaltracker.com/story/show/166266929


MKodde:
Looks good @pablothedude! Can we merge this?

PR #183, opened 2019-09-25
Use correct ResponseContext service identifier

MKodde:
Between regular SSO and SFO authentication, different response context
objects are used. It is very important that the correct one is used
(retrieved from container) during the ACS step. This was not the case
when dealing with SFO responses. In this commit, that is solved.

The problem is described in greater detail in the bug report:
https://www.pivotaltracker.com/story/show/168746898

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
@@ -261,11 +261,15 @@ class SecondFactorController extends Controller
             ));
         }
 
+        // Also send the response context service id, as later we need to know if this is regular SSO or SFO authn.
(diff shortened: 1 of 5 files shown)


epinxteren:
👍 Nice found the 🐛 ! 

I would strongly recommend that you add functional test for both cancel flows. 
(fix CI env)

MKodde:
We currently are not able to write functional tests for Stepup-Gateway except for platform wide behat tests in Stepup-Deploy. 

We have a dev-prio ticket for the next sprint to address this issue. https://www.pivotaltracker.com/story/show/166409479

MKodde:
In order to address the failing build, I've rebased the PR onto develop, which in the meantime received a Travis fixup (Ant is installed once again on the image).

PR #191, opened 2020-03-26
Create a SFO behat test

MKodde:
This is proof of concept. A yubikey authentication on the SFO endpoint is rigged up and tested. Happy flow only for now.

### DIFF
--- a/app/config/services_test.yml
+++ b/app/config/services_test.yml
@@ -16,6 +16,12 @@ services:
   surfnet_gateway_api.service.yubikey:
     class: Surfnet\StepupGateway\ApiBundle\Tests\TestDouble\Service\YubikeyService
 
+  gateway.service.gateway_api_sms:
+    class: Surfnet\StepupGateway\ApiBundle\Tests\TestDouble\Service\SmsService
+
+  surfnet_stepup.service.challenge_handler:
+    class: Surfnet\StepupGateway\ApiBundle\Tests\TestDouble\Service\SessionSmsVerificationStateHandler
+
   Surfnet\StepupGateway\Behat\MinkContext:
     class: Surfnet\StepupGateway\Behat\MinkContext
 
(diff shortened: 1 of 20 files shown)


app/config/services_test.yml
```diff
@@ -1,13 +1,40 @@
...
-  saml_sp_publickey: /vagrant/deploy/tests/behat/fixtures/test_public_key.crt
-  saml_sp_privatekey: /vagrant/deploy/tests/behat/fixtures/test_private_key.key
-  saml_idp_publickey: /vagrant/deploy/tests/behat/fixtures/test_public_key.crt
-  saml_idp_privatekey: /vagrant/deploy/tests/behat/fixtures/test_private_key.key
+  saml_metadata_publickey: /var/www/app/sp.crt
+  saml_metadata_privatekey: /var/www/app/sp.pem
+  saml_sp_publickey: /var/www/app/sp.crt
+  saml_sp_privatekey: /var/www/app/sp.pem
+  saml_idp_publickey: /var/www/app/idp.crt
+  saml_idp_privatekey: /var/www/app/idp.key
```
pablothedude, line 9: .pem / .key ?

ci/docker/init.sh
```diff
@@ -9,6 +9,7 @@ docker-compose up -d
 docker-compose exec -T php-fpm.stepup.example.com bash -c '
   cp ./ci/config/*.yml ./app/config/
```
pablothedude, line 10: Please remove the `cp` actions, only `composer up` should be sufficient to get this working to prevent configuration all over the place This script shouldn't manage state and this should be moved to service and parameter configuration. I've fixed this earlier on. It would even be better to even move the `composer` commands but this won't affect state so it could be left here to bootstrap the box. Ultimately this should be moved to the docker file itself.

pablothedude:
All work is already fixed so this PR could be closed

PR #192, opened 2020-04-02
Run selenium tests on Github Actions

MKodde:
Warning: this PR was constructed on top of `feature/create-sfo-behat-test`. That work is somewhat outdated with some of the changes introduced in this PR.

Notable changes:
1. mink-selenium2-driver was installed and configured to allow running of headless browser tests
1. Creating of SP's (and other db fixture data) now only inserts a new row if the entity did not yet exist. This is implemented using the PDO repositories. Not the cleanest code, could be improved later-on.
1. The existing consuemeAssertion twig templates have been updated. The value attribute was set on the submit-inputs. For some reason the auto submit does not work in the headless chrome. And debugging JS errors was outside of scope for now. (TODO).
1. A mock IdP (controller) was added. This is still a mess, but had to be implemented to also get regular SSO support in the behat features.
1. A new testing environemnt was introduced (thanks @pablothedude)
1. Some of the step definitions in the mink context needed


MKodde:
Changes from the child PR #193 have been merged into this PR.

pablothedude:
After functional testing I found some things that needs to be addressed. Feel free to offload it to different PR's or discuss any further. Let me know if I could be of any assistance.
- The used Node.js version in the Dockerfile is old and should be updated
- Not all tests in Deploy seem to work
- Fix app.php env to be used in all environments
- Enable and fix the `behat` and `behat-selenium` tests
- Fix Mopa assets
- Hardcode db hostname in service.yml config so it will work out of the box without changing the `parameters.yml`
- Use 'gateway_test' db instead of 'gateway' in service_(*).yml, maybe it would be even better to use//inject the DBAL connection?

pablothedude:
Are the points above already addressed?

MKodde:
**Sitrep**

The points above have either been  addressed in the now merged PR's. Or have been left open as a task on this Pivotal ticket: https://www.pivotaltracker.com/story/show/172199408

PR #195, opened 2020-04-09
Enable behat tests once again

MKodde:
To be able to run webtests and use the gateway in a dev/prod environment, the app.php was changed slightly to allow setting the Symfony environment using the APP_ENV environment variable. If that is not present on the _SERVER variables, prod is defaulted to.

This should not change current behaviour, but allow us to run web tests.

See https://www.pivotaltracker.com/story/show/172199408


pablothedude:
@MKodde do you know the status of this PR. Do the broken builds need to be fixed or is it safe to start a review?

MKodde:
@pablothedude Statusreport:
1. In order to run the selenium tests with the corrected Twig template, some selenium specific exceptions needed to be added to the Features.
2. Overriding the SF parameters using copy action in the Dockerfile was dropped in favour of setting the default parameters in the services_webtests.yml
3. I stopped running the builds on PUSH, this to limit build resources

PR #197, opened 2020-04-09
Run PHP 7.4 tests in parallel with existing 7.2

MKodde:
Separate Dockerfiles where created for the php fpm container. This
trickled down to a similar docker-compose and init script.

The test integration github actions file was updated to start both jobs


MKodde:
I have not found a Github Action instruction to mark a job as 'allowed to fail'. Is this an issue? Or do you see other options here @pablothedude ?

ci/docker/docker-compose-php74.yml
```diff
@@ -0,0 +1,62 @@
+version: '2.2'
+
+services:
```
pablothedude, line 3: I would suggest only add config to overide. So it's clear what is overridden. FYI: you can validate the end result with docker config

ci/docker/init-php74.sh
```diff
@@ -4,9 +4,9 @@ gid=$(id -g)
 
 printf "UID=${uid}\nGID=${gid}\nCOMPOSE_PROJECT_NAME=gateway" > .env
 
-docker-compose up -d --build
+docker-compose -f docker-compose-php74.yml up -d --build
```
pablothedude, line 7: Maybe this should be moved to the Github workspace config.

ci/docker/init-php74.sh
```diff
@@ -4,9 +4,9 @@ gid=$(id -g)
...
 
-docker-compose exec -T php-fpm.stepup.example.com bash -c '
+docker-compose -f docker-compose-php74.yml exec -T php-fpm.stepup.example.com bash -c '
```
pablothedude, line 9: Maybe this should be removed to the Dockerfile.

MKodde:
Yes, I will make note of your feedback, and let's address it in the near future

PR #198, opened 2020-04-14
Pass authentication context to error handling actions

MKodde:
This is a bugfix blocking the release of 17.1. The fix comprises of additional behat test coverage and the actual fix. The behat tests should be able to run without too much effort. But they will be properly installed for use on Github Actions in another PR.

More details about the bug can be found here: https://www.pivotaltracker.com/story/show/172262049


app/config/routing_test.yml
```diff
@@ -8,7 +8,6 @@ surfnet_stepup_gateway_behat_service_provider_acs:
 
 surfnet_stepup_gateway_behat_identity_provider_acs:
     path:     /test/authentication/sso
-    methods:  [GET]
```
pablothedude, line 11: Why not use 'GET' and 'POST' as defined in the prod routes?

tests/features/bootstrap/FeatureContext.php
```diff
@@ -85,6 +86,14 @@ public function aUserIdentifiedByWithAVettedToken($institution, $nameId, $tokenT
         }
     }
 
+    /**
+     * @Given /^a user from "([^"]*)" identified by "([^"]*)"$/
+     */
+    public function aUserIdentifiedBy($institution, $nameId)
+    {
+        // For now, do nothing, this user is not known to Gateway, but might exist.
```
pablothedude, line 94: If not used please remove. This only will make the Behat tests less readable if something states X will happen, but this isn't actually done.

tests/features/bootstrap/ServiceProviderContext.php
```diff
@@ -186,25 +212,49 @@ public function iStartAnAuthentication($nameId)
         $authnRequest->setDestination(self::SSO_ENDPOINT_URL);
         $authnRequest->setProtocolBinding(Constants::BINDING_HTTP_REDIRECT);
         $authnRequest->setNameId($this->buildNameId($nameId));
-        // Sign with random key, does not mather for now.
```
pablothedude, line 189: Why are these comments removed?

app/config/routing_test.yml
MKodde: Yeah, good point was just yanking stuff out, to get it to work to be honest. I will fix this right away.

tests/features/bootstrap/FeatureContext.php
MKodde: Yes, agreed, removing the step definition and its usages

tests/features/bootstrap/ServiceProviderContext.php
MKodde: Not sure why. I reverted that change.

PR #200, opened 2020-06-05
Remove usage of removed Mopa less

pablothedude:
Some Mopa leftover resulted in a exception
during frontend asset compilation.


MKodde:
Looking at my previous MOPA removal effort in https://github.com/OpenConext/Stepup-Gateway/pull/194. I think that PR was not successfully merged. Looking at the changeset of that PR shows me I fully removed the `bootstrap.less` file and modified the `style.less` somewhat. I see that my other changes from that branch have also been reset to a previous state. Maybe it would be best to merge the changes from 194 again?

PR #201, opened 2020-06-08
Cherrypick Mopa changes from #194

pablothedude:
Changes from #194 are cherry picked. This is needed to be
able top build the frontend assets successfully.


pablothedude:
I've tested the changes from the SA were it was cherrypicked from but this didn't work functionally. I think it's much better to use the changes from RA (using Encore in favour of Assetic) to get this working. This should be done when those components are getting upgraded so we could drop Assetic too. That way we also keep the steps to deploy the components the same.

Mopa was also needed implicitly for the language switcher by using the Stepup bundle and resulted in errors. It seems that the initial change in 194 wasn't tested thoroughly else this should have been detected earlier on. I'll close this PR, so this could be fixed in favour of Encore.

I'll close this PR.

See:
https://www.pivotaltracker.com/story/show/173959248
https://www.pivotaltracker.com/story/show/173959250

PR #204, opened 2020-11-19
Enforce Apache-2.0 license headers

mroest:


src/Surfnet/StepupGateway/GatewayBundle/Service/SecondFactorLocaleProvider.php
```diff
@@ -1,4 +1,19 @@
 <?php
+/**
+ * Copyright 2020 SURFnet bv
```
MKodde, line 3: Nitpick: when adding new copyright headers, please use the correct `B.V.` version.

MKodde (approved):
Awesome! Consider applying the proposed change, or merge as is. Both outcomes are fine with me.

PR #209, opened 2020-12-11
WIP: Replace Messagebird with Spryng

phavekes:


src/Surfnet/StepupGateway/ApiBundle/Service/SmsService.php
```diff
@@ -56,13 +48,24 @@ public function send(SmsMessage $message, Requester $requester)
     {
         $this->logger->notice('Sending OTP per SMS.');
 
-        $message = new Message($message->originator, $message->recipient, $message->body);
-        $result = $this->messagingService->send($message);
+        // TODO: get apikey from parameters file
+        $spryng = new Spryng("dummy");
```
MKodde, line 52: As a proof of concept this will work very well. For it to match our somewhat layered infrastructure. I'd prefer a SpryngClient that is configured with the api key via dependency injection. That client could well be the MessagingService (which should be an interface) a new implementation of that interface could be the SpryngService. That way other StepUp users can keep using MessageBird.

If you'd like me to give you a nudge in the right direction, I'd be more than happy to help.

phavekes:
Will be solved in https://github.com/OpenConext/Stepup-Gateway/pull/221

PR #213, opened 2021-02-22
Support for SAML extensions

mroest:
This will enable passing on user attributes on the saml
message.

See: https://www.pivotaltracker.com/story/show/175121584


MKodde (commented):
Nice and simple! I like!

Maybe updating the Stepup Saml bundle composer constraint should be part of this PR? If you ask me, https://github.com/OpenConext/Stepup-saml-bundle/pull/101 is ready for release. :rocket:

PR #218, opened 2021-04-26
Fix authentication logging 

phavekes:
R21 removed auth-logging

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -59,10 +59,6 @@
                 "assertion",
                 "validation"
             ],
-            "support": {
-                "issues": "https://github.com/beberlei/assert/issues",
-                "source": "https://github.com/beberlei/assert/tree/v2.9.9"
-            },
             "time": "2019-05-28T15:27:37+00:00"
         },
         {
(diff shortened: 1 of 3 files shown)


MKodde:
The full problem is not fully clear to me. Did the authentication logs end up in every log? And was that not meant to be? Reading your change, it is now excluded from the prod-signaler and the syslog and the auth log now only get the the authentication channel. But given the somewhat complex ways Monolog is configured, I might be mistaken?

Can you elaborate?

phavekes:
> The full problem is not fully clear to me. Did the authentication logs end up in every log? And was that not meant to be? Reading your change, it is now excluded from the prod-signaler and the syslog and the auth log now only get the the authentication channel. But given the somewhat complex ways Monolog is configured, I might be mistaken?
> 
> Can you elaborate?

Yes, the authentication-logging needs to end up in a different format in a different location. Before the php7/sf4 migration this worked fine, But during migration the logging config was removed from 
https://github.com/OpenConext/Stepup-Gateway/blob/release-17.1/app/config/logging.yml, 
and added in config/packages/monolog.yaml, but the filter was removed.

MKodde:
Thanks this clarifies the situation. And is also useful for future reference.

MKodde (approved):

PR #220, opened 2021-05-03
Provide Behat test support

MKodde:
Symfony is kicked into smoketest mode when the testcookie is set.


Badlapje (approved):
Not sure if i am the right person to review this as this really isn't my cup of tea.  That said, i see nothing funky going on.  So presumably a great job 😝!

MKodde:
Given this is a low impact change, I feel happy merging this. And don't underestimate yourself! You're quickly becoming a SF Jedi :zap:

PR #234, opened 2022-01-14
Delete u2f remnant app-id

thijskh:
Seems to have been accidentally (re)added in #221, but u2f support is no longer present in sa-gw. Also the web/ dir is no longer served.


thijskh:
It seems that the build fails on the security checker only.

phavekes (approved):
Nice find @thijskh. Thanks for correcting this.

PR #235, opened 2022-01-14
Add missing Disallow line to robots.txt

thijskh:
Robots.txt specifed a list of User-Agents (*) but not actions for these
agents which makes no sense. The probable intended action is Disallow
since the Stepup Gateway is an API, not something web crawlers should
index. For reference, engineblock also has "Disallow: /"
This saves wasted resources and noise in the logs.


phavekes (approved):
Thanks for spotting and correcting this!

PR #241, opened 2022-02-14
Use internal-collabPersonId when present

MKodde:
When EB sends the attribute urn:mace:surf.nl:attribute-def:internal-collabPersonId:
* Use this attribute's value as the person's identifier Stepup uses internally.
* Remove this attribute from the set of attributes. It must not be released to the SP in any situation,
* Use the supplied NameID as the NameID to the SP
* Treat the epTID not specially when present. If it's present it will be released to the SP just like another attribute that would be present.

Otherwise, when the attribute is not set:
* Keep current behaviour = Use NameID as person's identifier ,Use epTID as value for NameID for SP, 

Story: https://www.pivotaltracker.com/story/show/181115261

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Service/ProxyResponseService.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Service/ProxyResponseService.php
@@ -55,11 +55,6 @@ class ProxyResponseService
     private $attributeDictionary;
 
     /**
-     * @var \Surfnet\SamlBundle\SAML2\Attribute\AttributeDefinition
-     */
-    private $eptiAttribute;
-
-    /**
      * @var \DateTime
      */
     private $currentTime;
@@ -79,7 +74,6 @@ class ProxyResponseService
         ProxyStateHandler $proxyStateHandler,
         AssertionSigningService $assertionSigningService,
         AttributeDictionary $attributeDictionary,
-        AttributeDefinition $eptiAttribute,
         Loa $intrinsicLoa,
         DateTime $now = null
     ) {
@@ -87,7 +81,6 @@ class ProxyResponseService
         $this->proxyStateHandler         = $proxyStateHandler;
         $this->assertionSigningService   = $assertionSigningService;
         $this->attributeDictionary       = $attributeDictionary;
-        $this->eptiAttribute             = $eptiAttribute;
         $this->intrinsicLoa              = $intrinsicLoa;
         $this->currentTime = is_null($now) ? new DateTime('now', new DateTimeZone('UTC')): $now;
     }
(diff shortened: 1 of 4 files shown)


Badlapje (approved):
I can't test this functionally, but code-wise this looks great to me.  Here be tests, and code which looks good.  Nice work!

src/Surfnet/StepupGateway/GatewayBundle/Tests/Service/ProxyResponseServiceTest.php
```diff
@@ -155,10 +214,10 @@ public function it_appends_the_proxied_identity_provider_to_existing_authenticat
         );
     }
 
-    public function testCreateProxyResponseRequiresEpti()
+    public function testCreateProxyResponseRequiresEptiIfInternalCollabPersonIdIsNotPresent()
     {
         $this->expectException(RuntimeException::class);
-        $this->expectExceptionMessage('The "urn:mace:dir:attribute-def:eduPersonTargetedID" is not present');
+        $this->expectExceptionMessage('Neither "urn:mace:dir:attribute-def:eduPersonTargetedID" or "urn:mace:surf.nl:attribute-def:internal-collabPersonId" is present');
```
pmeulen, line 220: "Neither" goes with "nor"

src/Surfnet/StepupGateway/GatewayBundle/Service/ProxyResponseService.php
```diff
@@ -110,21 +110,17 @@ public function createProxyResponse(Assertion $assertion, $destination, $loa = n
...
-        // Perform some input validation on the eptiNameId that was received.
-        if (is_null($eptiNameId)) {
-            throw new RuntimeException('The "urn:mace:dir:attribute-def:eduPersonTargetedID" is not present.');
-        } elseif (!array_key_exists(0, $eptiNameId) || !$eptiNameId[0]->value) {
+        $internalCollabPersonId = $translatedAssertion->getAttributeValue('internalCollabPersonId');
+        // Perform some input validation on the eptiNameId and/or internal-collabPersonId that was received.
+        if (is_null($eptiNameId) && is_null($internalCollabPersonId)) {
             throw new RuntimeException(
-                'The "urn:mace:dir:attribute-def:eduPersonTargetedID" attribute does not contain a NameID with a value.'
+                'Neither "urn:mace:dir:attribute-def:eduPersonTargetedID" or ' .
```
pmeulen, line 119: Typo: "or" => "nor"

pmeulen (changes requested):
Logic looks sound and tests exist.

One small typo to fix.

PR #242, opened 2022-02-23
Add a slash to the .idea .gitignore entry

MKodde:
Git will still see the idea folder as a stagable entity if we do not make explicit it is a folder


VadimSchmitz (approved):

VadimSchmitz:
travis ci build broke on symphony webpack due to a vulnerability.

PR #248, opened 2022-04-12
Resolve GSSP metadata generation twig environment configuration issue

VadimSchmitz:
Work done in this pull request:

- The local service `second_factor_only.metadata_factory` has now been removed so it uses the `Stepup-saml-bundle` configuration from the Stepup Saml Bundle.
- Update `SurfnetStepupGatewaySamlStepupProviderExtension.php` to use the new rendering method `twig` instead of the no longer available `templating`
- Update `ResponseRenderingService.php`

https://www.pivotaltracker.com/story/show/181710432

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Service/ResponseRenderingService.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Service/ResponseRenderingService.php
@@ -22,7 +22,7 @@ use SAML2\Constants;
 use SAML2\Response as SAMLResponse;
 use Surfnet\StepupGateway\GatewayBundle\Saml\ResponseBuilder;
 use Surfnet\StepupGateway\GatewayBundle\Saml\ResponseContext;
-use Symfony\Bundle\TwigBundle\TwigEngine;
+use Twig\Environment;
 use Symfony\Component\HttpFoundation\Response;
 
 final class ResponseRenderingService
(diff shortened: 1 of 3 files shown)


MKodde (changes requested):
Thanks for fixing the bug! I have some feedback:
1. You have introduced several formatting issues. That's what the PHP CS is triggering on I recon.
2. Please reword the first commit. Using the 'fix this and that' can always be rewritten to be more specific.
3. XML output looks great again!
4. I reworded your PR message and title slightly. Making it easier to read.

VadimSchmitz:
The static-analysis tests currently pass, however the build currently fails due to a vulnerability warning.

src/Surfnet/StepupGateway/GatewayBundle/Service/ResponseRenderingService.php
```diff
@@ -74,14 +74,14 @@ public function renderRequesterFailureResponse(ResponseContext $context)
...
-        return $this->renderSamlResponse(
+        return ($this->renderSamlResponse(
             $context,
             'unprocessable_response',
             $this->responseBuilder
                 ->createNewResponse($context)
                 ->setResponseStatus(Constants::STATUS_RESPONDER)
                 ->get()
-        );
+        ));
```
MKodde, line 84: Wrapping the return statement in `(` brackets `)` is not adding any value right?

src/Surfnet/StepupGateway/GatewayBundle/Service/ResponseRenderingService.php
```diff
@@ -104,16 +104,18 @@ public function renderResponse(
      */
     private function renderSamlResponse(
         ResponseContext $context,
-        $view,
-        SAMLResponse $response
+        string          $view,
```
MKodde, line 107: :+1: for adding the type def!

MKodde (commented):
One change should be reverted. The rest looks great now!

MKodde (approved):
Grade A job!

:ship:

PR #258, opened 2022-07-06
Support self-asserted LoA

MKodde:
Gateway is now able to handle step up (regular and SFO) authentications with self-asserted tokens. These tokens have a different level of assurance opposed to tokens that are vetted with any of the other vetting types. The LoA for these tokens is lowered to a level of 1.5. Not due to the strength of the token, but because of the strength of the registration procedure used.

Gateway uses the updated Stepup-bundle to ascertain the correct LoA for any given second factor token.

Behat tests where added for ascertaining the correct behavior in Gateway. Some tweaking was required to get the tests running.

See: https://www.pivotaltracker.com/story/show/181933334
See: https://github.com/OpenConext/Stepup-SelfService/pull/265


tests/features/bootstrap/MinkContext.php
```diff
@@ -79,6 +79,7 @@ public function theResponseShouldMatchXpath($xpath)
      */
     public function theResponseShouldNotMatchXpath($xpath)
     {
+        $this->printLastResponse(); die;
```
pablothedude, line 82: Should this `die` be kept here?

tests/features/bootstrap/ServiceProviderContext.php
```diff
@@ -166,14 +167,48 @@ public function iStartAnSFOAuthentication($nameId)
             $this->loadPrivateKey(new PrivateKey('/var/www/ci/certificates/sp.pem', 'default'))
         );
         $authnRequest->setRequestedAuthnContext(
-            ['AuthnContextClassRef' => ['http://stepup.example.com/assurance/sfo-level2']]
+            ['AuthnContextClassRef' => ['http://stepup.example.com/assurance/loa-self-asserted']]
```
pablothedude, line 170: Does the name of this method reflect the authnclassref used? Or is this deliberate?

tests/features/self-asserted.feature
```diff
@@ -0,0 +1,25 @@
+Feature: As an institution that uses the sefl-asserted tokens (SAT) registration
```
pablothedude, line 1: :100:

pablothedude (approved):

PR #259, opened 2022-07-08
Log when proxy state handler is missing

phavekes:


src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
```diff
@@ -569,9 +569,15 @@ private function getResponseContext($authenticationMode)
     {
         switch ($authenticationMode) {
             case self::MODE_SFO:
+                if (!($this->get('gateway.proxy.sfo.state_handler'))) {
+                    throw new RuntimeException('Cannot find SFO state handler.');
+                }
```
MKodde, line 574: 1. Was this check added based on actual 500 errors ending up in the log?
2. What is the benefit of this RuntimeException over the harder error that was already thrown from the container?

And would it not be more plausible that the `getResponseContextServiceId` service would be unavailable on the container? Chances are very slim that the sso/sfo state_handlers would not be present.

All this aside. The best solution here would be to utilize dependency injection over fishing out services from the service container.. Work for another day I suppose.

phavekes: It would indeed be better to just log this. Changes made

PR #264, opened 2022-11-03
Read institution configuration projection

MKodde:
See task 3 of: https://www.pivotaltracker.com/story/show/183402518


src/Surfnet/StepupGateway/GatewayBundle/Entity/InstitutionConfiguration.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+/**
+ * Copyright 2016 SURFnet B.V.
```
Shaky212, line 4: Seems a new file, update year to 2022 and  use strict_types?

Shaky212 (commented):
Just some nitpicking about the header.

PR #270, opened 2022-12-06
Configure exceptions to Allow|Store SSO on 2FA

MKodde:
Don't store the SSO cookie when a RA vetting procedure is handled by Gateway.

Don't try to give SSO on 2FA when one of the `gssp_allowed_sps` is used during authentication.

See: https://www.pivotaltracker.com/story/show/183511012


src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
```diff
@@ -20,6 +20,7 @@
 
 use Surfnet\StepupGateway\GatewayBundle\Saml\Exception\RuntimeException;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use function is_bool;
```
Shaky212, line 23: Seemed to be never used.

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Provider/ConnectedServiceProviders.php
```diff
@@ -18,67 +18,47 @@
...
     }
 
-    /**
-     * @param string $serviceProvider
-     * @return bool
-     */
-    public function isConnected($serviceProvider)
+    public function isConnected(string $serviceProvider): bool
     {
         if (!is_string($serviceProvider)) {
```
Shaky212, line 46: serviceProvider is passed by as a string, seems that this condition is always false?

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Provider/ConnectedServiceProviders.php
```diff
@@ -18,67 +18,47 @@
...
     }
 
-    /**
-     * @param string $serviceProvider
-     * @return \Surfnet\StepupGateway\GatewayBundle\Entity\ServiceProvider
-     */
-    public function getConfigurationOf($serviceProvider)
+    public function getConfigurationOf(string $serviceProvider): ServiceProvider
     {
         if (!is_string($serviceProvider)) {
```
Shaky212, line 54: serviceProvider is passed by as a string, seems that this condition is always false?

Shaky212 (commented):
Just a few minor comments

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Provider/ConnectedServiceProviders.php
MKodde: Good point, I removed the obsolete input validation checks!

PR #271, opened 2022-12-06
Log SSO on 2FA authentications to auth-log

MKodde:
To get the correct data into the log message. Some additional stowing of data was required on the context (state handler).

See: https://www.pivotaltracker.com/story/show/183511341


Shaky212 (approved):
Looks good!

PR #272, opened 2022-12-19
Store resulting loa, not the required

MKodde:
The required (required by SP) LoA was stored in the SSO cookie. That is not rigth, we want to store the LoA of the SF used to authenticate the rquired LoA. That LoA might be higher. An might assist the end user in not having to give another 2FA when another SP requires a higher LoA.

See: https://www.pivotaltracker.com/story/show/184060757
See: https://www.pivotaltracker.com/story/show/183402542


Shaky212 (approved):
Looks good!

PR #273, opened 2022-12-21
Upgrade Webpack build tools

MKodde:
Webpack encore and its build tooling was upgraded to the latest version


Shaky212 (approved):
Looks good!

PR #279, opened 2023-03-30
Add ADFS parameters to SAMLResponse on error

MKodde:
When a response discloses a saml failed response, the current setup would not add the Context and AuthMethod to the POST values that are sent back to the ADFS SP.

By adding these ADFS params for any SAMLResponse the ADFS sp will be able to handle the error response.

Some knowledge that migth aid in reviewing this PR:
1. The ADFS SAMLResponse POST parameter is prepended with an underscore. This is to serve the ADFS plugin that picks up the SAML response after GW is done with it.
2. ADFS utilizes Post binding for both sending the AuthNRequest, and for receiving back the SAMLResponse
3. Previously when an ADFS authentication was cancelled, or it failed for some other reason. The ADFS parameters where not posted back, resulting in the fact that the ADFS plugin could not further process the error response.
4. Behat tests where added. Simulating the ADFS flow. The Context field would normally be filled with actual data. In our case an empty `<EncryptedData>` element is transported.
5. The Behat SFO tests where once again added to the non-selenium set of behat tests. This did need some aditional tweaking and refactoring. So some of the content of the PR is somewhat unrelated to the actual change. Consider this boyscout work ;)

See https://www.pivotaltracker.com/story/show/184314508

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
@@ -28,6 +28,7 @@ use Surfnet\StepupGateway\GatewayBundle\Service\Gateway\ConsumeAssertionService;
 use Surfnet\StepupGateway\GatewayBundle\Service\Gateway\FailedResponseService;
 use Surfnet\StepupGateway\GatewayBundle\Service\Gateway\LoginService;
 use Surfnet\StepupGateway\GatewayBundle\Service\Gateway\RespondService;
+use Surfnet\StepupGateway\SecondFactorOnlyBundle\Adfs\ResponseHelper;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -191,14 +192,33 @@ class GatewayController extends Controller
      */
     public function renderSamlResponse($view, SAMLResponse $response, $authenticationMode)
     {
+        $logger = $this->get('logger');
+        /** @var ResponseHelper $responseHelper */
+        $responseHelper = $this->get('second_factor_only.adfs.response_helper');
+
         $this->supportsAuthenticationMode($authenticationMode);
         $responseContext = $this->getResponseContext($authenticationMode);
 
-        return $this->render($view, [
-            'acu'        => $responseContext->getDestination(),
-            'response'   => $this->getResponseAsXML($response),
+        $parameters = [
+            'acu' => $responseContext->getDestination(),
+            'response' => $this->getResponseAsXML($response),
             'relayState' => $responseContext->getRelayState()
-        ]);
+        ];
+
+        // Test if we should add ADFS response parameters
+        $inResponseTo = $responseContext->getInResponseTo();
+        if ($responseHelper->isAdfsResponse($inResponseTo)) {
+            $adfsParameters = $responseHelper->retrieveAdfsParameters();
+            $logMessage = 'Responding with additional ADFS parameters, in response to request: "%s", with view: "%s"';
+            if ($response->isSuccess()) {
+                $logMessage = 'Responding with an AuthnFailed SamlResponse with ADFS parameters, in response to AR: "%s", with view: "%s"';
+            }
+            $logger->notice(sprintf($logMessage, $inResponseTo, $view));
+            $parameters['adfs'] = $adfsParameters;
+            $parameters['acu'] = $responseContext->getDestinationForAdfs();
+        }
+
+        return $this->render($view, $parameters);
     }
 
     /**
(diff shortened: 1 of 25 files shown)


src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/Gateway/LoginService.php
```diff
@@ -128,6 +128,9 @@ public function singleSignOn(Request $httpRequest, ReceivedAuthnRequest $origina
 
         // Check if the NameID is provided and we may use it.
         $nameId = $originalRequest->getNameId();
+        if (!$nameId) {
+            throw new RequesterFailureException('The NameId was not present in the AuthNRequest');
+        }
```
MKodde, line 133: This change seems unrelated to the story, and it is. 

During testing of the simulated ADFS authnrequests/samlresponse I ran into a situation where the nameId was not present in the AuthNRequest. And that resulted in an uncaught error. I added this `RequesterFailureException` to get a somewhat more detailed error report in the logs.

/boyscout rule

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
```diff
@@ -140,8 +140,7 @@ public function respondAction()
                 [
                     'acu' => $responseContext->getDestinationForAdfs(),
                     'samlResponse' => $xmlResponse,
-                    'context' => $adfsParameters->getContext(),
-                    'authMethod' => $adfsParameters->getAuthMethod(),
+                    'adfs' => $adfsParameters,
```
MKodde, line 143: To use the adfs partial, the existing SFO controller action that handled the adfs response, no sends the adfs parameters value object directly to the view.  The view tests if the parameters are present, and adds them to the post params.

src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
```diff
@@ -191,14 +192,33 @@ public function sendAuthenticationCancelledByUserAction()
...
+            $logMessage = 'Responding with additional ADFS parameters, in response to request: "%s", with view: "%s"';
+            if ($response->isSuccess()) {
+                $logMessage = 'Responding with an AuthnFailed SamlResponse with ADFS parameters, in response to AR: "%s", with view: "%s"';
+            }
+            $logger->notice(sprintf($logMessage, $inResponseTo, $view));
+            $parameters['adfs'] = $adfsParameters;
+            $parameters['acu'] = $responseContext->getDestinationForAdfs();
+        }
+
+        return $this->render($view, $parameters);
```
MKodde, line 221: This might seem illogical. We are responding to an SFO SAML Response (Adfs authentications are intrinsically SFO auth) from the 'non-sfo' Gateway controller. 

That is because all AuthnFailed saml responses are redirected via that route. And that is why those endpoints now also support the ADFS response parameters.

Not my favourite, and feel free to discuss a possible better solution. For one, I reconsidered adding an Event Listener on the response sent event. We could add the additional ADFS parameters there for any ACS action that is performed. But that also raises the magic factor up a notch..

tests/features/sfo.feature
```diff
@@ -12,14 +11,15 @@ Feature: As an institution that uses the second factor only feature
...
 
-  Scenario: A SMS SFO authentication
-    Given an SFO enabled SP with EntityID https://sp.stepup.example.com
-    And a whitelisted institution stepup.example.com
-    And a user from "stepup.example.com" identified by "urn:collab:person:stepup.example.com:blaine_sumner" with a vetted "SMS" token
-    When urn:collab:person:stepup.example.com:blaine_sumner starts an SFO authentication
-    Then I should see the SMS verification screen
-    When I enter the SMS verification code
-    Then the response should match xpath '//samlp:StatusCode[@Value="urn:oasis:names:tc:SAML:2.0:status:Success"]'
+# TODO: SMS verification does not work. The hard coded `432543` SMS secret is not accepted the mock SMS service, is it not in place?
```
pablothedude, line 14: Is this still work in progress?

src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
pablothedude: Maybe consider a dedicated failed SAML response route to redirect to to handle failed authentications. Or is that not possible?
Not raising the level of magic however is certainly a good call!

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/Gateway/LoginService.php
pablothedude: :100: and thanks for the heads up!

tests/features/adfs.feature
```diff
@@ -0,0 +1,31 @@
+Feature: As an institution that uses ADFS support on the second factor only feature
```
pablothedude, line 1: :100:

pablothedude (approved):

tests/features/sfo.feature
MKodde: No, I disabled this scenario as it was the only one of the SFO tests not working correctly at the moment. I will fix this at some point in the future.

PR #283, opened 2023-07-04
Set response context service id on state handler

MKodde:
When starting a GSSP registration, the response context service id was not set on the state handler. Causing the new improved error response redirection to fail. As the state handler would not be able to load the correct response context service to get the Issuer entity id from. Causing the response to end up in a 500 error, and not a 301 back to the issuer.


.github/workflows/daily-security-check.yml
```diff
@@ -0,0 +1,84 @@
...
+        if: steps.check_node_npm.outputs.files_exists == 'true'
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14
+      - name: npm audit
+        if: steps.check_node_npm.outputs.files_exists == 'true'
+        run: npm audit --audit-level=high
+        continue-on-error: true
+
+      # python checks
```
pablothedude, line 59: I was wondering for what the Python and Java tests are used?

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
```diff
@@ -243,11 +245,19 @@ private function getIssuer(StateHandler $stateHandler): Issuer
     {
         // This can either be a SFO or 'regular' SSO authentication. Both use a ResponseContext service of their own
         $responseContextServiceId = $stateHandler->getResponseContextServiceId();
+        if (!$responseContextServiceId) {
```
pablothedude, line 248: There are relatively often problems with the `StateHandler` and  maybe you could opt to refactor it to be more robust to prevent bugs in the future.

pablothedude (approved):

.github/workflows/daily-security-check.yml
MKodde: This very generic script is used by other SURF project. And is also compatible with Java and Python apps. Leaving us with a predictable way of getting security reports for all kinds of projects.

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Controller/SamlProxyController.php
MKodde: I'm very reluctant to update it here and now. But yes you are right. The state handlers often cause confusion. Your comment did cause me to update the `gateway-state-gssp-registration-flow.png` diagram. Which aids in showing developers how/when state is used and for what purpose.

Issue #286, opened 2023-07-24
ADFS Plugin specific response format not used SAML Error from GSSP


pmeulen:
in https://github.com/OpenConext/Stepup-Gateway/tree/4.1.2, when a GSSP returns a SAML Response with an error status, the response format required by the SFO ADFS extension is not used. This case was missed when fixing this in #285 and #279

There are three views called `consume_assertion.html.twig` in the Stepup-Gateway codebase:

1. https://github.com/OpenConext/Stepup-Gateway/blob/4.1.2/src/Surfnet/StepupGateway/GatewayBundle/Resources/views/gateway/consume_assertion.html.twig
2. https://github.com/OpenConext/Stepup-Gateway/blob/4.1.2/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Resources/views/saml_proxy/consume_assertion.html.twig
3. https://github.com/OpenConext/Stepup-Gateway/blob/4.1.2/src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Resources/views/adfs/consume_assertion.html.twig

The 2nd one (https://github.com/OpenConext/Stepup-Gateway/blob/65f5bb6a8adb05cd23995891f250d89eadf07e57/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Resources/views/saml_proxy/consume_assertion.html.twig) from the SamlStepupProviderBundle is used to render the error response from a GSSP, and it does not contain the ADFS handling. The first one, from the GatewayBundle, does contain the ADFS handing. The 3rd one, from the SecondFactorOnlyBundle, is specifically for rendering an ADFS response, it cannot render a normal SAML Response.

Since the previous PR made `adfs.context` and `adfs.authMethod` available to the view from the SamlStepupProviderBundle, ADFS specific rending can be added in the same way as was done in the `consume_assertion` view from the GatewayBundle.

PR #287, opened 2023-07-24
Fix for #286: ADFS SFO plugin specific parameters are not returned on GSSP error

pmeulen:
See #286

The question is why these two different templates are used.

### DIFF
--- a/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Resources/views/saml_proxy/consume_assertion.html.twig
+++ b/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Resources/views/saml_proxy/consume_assertion.html.twig
@@ -3,15 +3,22 @@
 {% block page_title %}{{ 'gateway.gssp.consume_assertion.title'|trans }}{% endblock %}
 
 {% block content %}
+    {% set samlResponse = 'SAMLResponse' %}
+    {% if adfs is defined %}
+        {# When responding to the ADFS plugin, we add an underscore to the `SAMLResponse` parameter #}
+        {% set samlResponse = '_SAMLResponse' %}
+    {% endif %}
     <noscript>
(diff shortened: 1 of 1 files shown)


MKodde (approved):
The change makes sense! And this will resolve the issue most certainly.

To answer you question

> The question is why these two different templates are used.

Each proxy feature (SFO, GSSP, SSO) has a consume assertion twig template of it's own. This is mostly dictated by the Symfony bundle principle. Where a bundle can be seen as an isolated part of the application. So has templates of its own. 

I agree, in this case it's not very practical.

Suggestions for the backlog: 
- We could introduce a global `consume_assertion` Twig template
- Keep three `consume_assertion` templates, but make them use global partials to extract the shared logic

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Resources/views/saml_proxy/consume_assertion.html.twig
```diff
@@ -3,15 +3,22 @@
...
     </noscript>
     {{ 'gateway.gssp.consume_assertion.one_moment_please'|trans }}
     <form method="post" action="{{ acu }}">
-        <input type="hidden" name="SAMLResponse" value="{{ response }}" />
+        <input type="hidden" name="{{ samlResponse }}" value="{{ response }}" />
     {% if relayState|length > 0 %}
         <input type="hidden" name="RelayState" value="{{ relayState|escape }}" />
     {% endif %}
+    {# Optionally add the ADFS response parameters #}
+    {% include "@SurfnetStepupGatewaySecondFactorOnly/adfs/partial/adfs.html.twig" %}
```
MKodde, line 21: It would be golden to move this partial to the `webroot/templates/partial` folder. As it is now used in two bundles.

PR #290, opened 2023-08-09
Resolve SF deprecation warnings

MKodde:
Quite some warnings where resolved. The Behat output (see QA run on GHA) is now almost cleaned up. The one remaining warning there is the one triggered via the JS Translation Bundle:

`User Deprecated: The Symfony\Bundle\TwigBundle\Loader\FilesystemLoader class is deprecated since version 4.3 and will be removed in 5.0; use Twig notation for templates instead.`

We could rid ourself of that by simply upgrading the bundle, but that requires us to also upgrade to PHP8. So let's skip that for now.


.scrutinizer.yml
```diff
@@ -1,7 +1,7 @@
 filter:
     excluded_paths:
         - "*/Tests/*"
-        - "app/DoctrineMigrations/*"
```
johanib, line 4: Is this not included in the previous line? Or is it case sensitive?

src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
```diff
@@ -19,8 +19,8 @@
 namespace Surfnet\StepupGateway\GatewayBundle\Saml\Proxy;
 
 use Surfnet\StepupGateway\GatewayBundle\Saml\Exception\RuntimeException;
+use Symfony\Component\HttpFoundation\Session\SessionBagInterface;
```
johanib, line 22: SessionBagInterface is not used in this file? Why is an import needed?

If this is really an error, consider tooling that removes unused imports.

johanib (approved):
few remarks, but no functional issues afaics

src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
MKodde: Thanks, removed that import

.scrutinizer.yml
MKodde: Thanks, I was mislead by the removed line. I later found out that the newly added line was not having the effect I looked for. And I'm going to remove it once again. But your feedback is interesting, I do not think the excluded path regex is not performed in a case insensitive manner.

PR #291, opened 2023-08-10
Repair the Selenium Behat tests

MKodde:
The SFO multiple authentications feature was disabled a while ago. They have been re-instated in this PR

Test one and two have been fixed. The Yubikey and SMS tests could be enabled in the new (SSP based) test environment. The Tiqr variant could not yet be enabled as Tiqr is not functioning in that environment yet.

During enabling the SMS test, I found out that the SMS Bypass fix was not taken into account in the existing sfo test. That has been updated.

Commit 2 consists of a boyscout rule fix. 

@pmeulen and or @phavekes 

1. Is this indeed the case at this point (does the WAYG page have an empty title?)
2. Using the previously used title seemed the right thing to do. 

We might not want to change this behavior for the 4.2 release?


public/index.php
```diff
@@ -23,7 +23,10 @@
 // To run behat tests in smoketest mode, the app env needs to be 'dev' or 'test'
 // and the user agent needs to be that of the behat guzzle client.
 $isTestOrDev = ($_SERVER['APP_ENV'] === 'dev' || $_SERVER['APP_ENV'] === 'test');
-if ($isTestOrDev && $_SERVER['HTTP_USER_AGENT'] === 'Symfony BrowserKit') {
+// When running the selenium tests, a test cookie indicate if we should switch to the smoketest env
+$hasTestCookie = (array_key_exists('testcookie', $_COOKIE) && $_COOKIE['testcookie'] === 'testcookie');
+
+if (($isTestOrDev && $_SERVER['HTTP_USER_AGENT'] === 'Symfony BrowserKit') || ($isTestOrDev && $hasTestCookie)) {
```
johanib, line 29:
```suggestion
if ($isTestOrDev && ($_SERVER['HTTP_USER_AGENT'] === 'Symfony BrowserKit' || $hasTestCookie)) {
```

tests/features/bootstrap/MinkContext.php
```diff
@@ -164,6 +168,8 @@ public function iOpenTwoBrowserTabsIdentifiedBy($numberOfTabs, $tabNames)
                 'Please identify all tabs you are opening in order to refer to them at a later stage'
             );
         }
+        // Set the testcookie ensuring the selenium tests run in the smoketest env
+        $this->getMink()->getSession()->setCookie('testcookie', 'testcookie');
```
johanib, line 172: is it possible to use more descriptive naming? For example: 'isBehatTest'

johanib (approved):

tests/features/bootstrap/MinkContext.php
MKodde: A good suggestion, but I choose to adhere to the way we called this cookie in OpenConext-engineblock. That way, we have feature parity between the projects.

public/index.php
MKodde: Yes, that is more betterest, going to fix that up!

PR #293, opened 2023-08-16
Verify the SSO cookie expiration date

MKodde:
The cookie lifetime can be extended by the user. But the explicit lifetime of the cookie was not yet validated by the cookie service. Meaning you could extend the SSO lifetime by keeping your cookie alive.

As of this change, the authentication lifetime is checked against the current timestamp togegher with the cookie expiration time.

See: https://www.pivotaltracker.com/story/show/183402734
> the SSO cookie's timestamp is within the configured SSO lifetime

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Sso2fa/DateTime/ExpirationHelper.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Sso2fa/DateTime/ExpirationHelper.php
@@ -32,15 +32,23 @@ class ExpirationHelper implements ExpirationHelperInterface
     private $now;
 
     /*
+     * The period in seconds that we still acknowledge the
+     * cookie even tho the expiration was reached. This accounts
+     * for server time/sync differences that may occur.
+     */
+    private $gracePeriod;
+
+    /*
      * The SSO on 2FA cookie lifetime in seconds
      *
      * See: config/legacy/parameters.yaml sso_cookie_lifetime
      */
     private $cookieLifetime;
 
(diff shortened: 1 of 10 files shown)


src/Surfnet/StepupGateway/GatewayBundle/Sso2fa/DateTime/ExpirationHelper.php
```diff
@@ -0,0 +1,86 @@
...
+    private $gracePeriod;
+
+    /*
+     * The SSO on 2FA cookie lifetime in seconds
+     *
+     * See: config/legacy/parameters.yaml sso_cookie_lifetime
+     */
+    private $cookieLifetime;
+
+    public function __construct(int $cookieLifetime, int $gracePeriod, CoreDateTime $now = null)
```
johanib, line 48: Waarom een alias CoreDateTime ipv gewoon DateTime?

johanib (approved):
Heb je ook geredeneerd dat het goed gaat met bestaande cookies die deze waarde nog niet hebben?

Verder ziet het er goed uit

src/Surfnet/StepupGateway/GatewayBundle/Sso2fa/DateTime/ExpirationHelper.php
MKodde: I tried my best to use the domain specific DateTime implementation. Which produces a /DateTime, which I aliassed to CoreDateTime as our linting discourages using unimported classes.

PR #295, opened 2023-09-04
Address some remaining SSO on 2FA issues

MKodde:
1. The resetting of the 2FA sso verification state was moved to the response sent action. The caused the removal of the setVerifiedBySsoOn2faCookie method. But the GSSP verified action still called it
2. The SFO authentication did not yet set the SSO cookie (when applicable)
3. The final 2 commits deal with 2 smaller issues.

For tests, see: https://www.pivotaltracker.com/story/show/185989315

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Service/GlobalViewParameters.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Service/GlobalViewParameters.php
@@ -55,8 +55,12 @@ final class GlobalViewParameters
     /**
      * @return string
      */
-    public function getSupportUrl()
+    public function getSupportUrl(): string
     {
-        return $this->supportUrl[$this->translator->getLocale()];
+        $locale = $this->translator->getLocale();
+        if (array_key_exists($locale, $this->supportUrl)) {
+            return $this->supportUrl[$locale];
+        }
+        return '';
     }
 }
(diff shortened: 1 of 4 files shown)


phavekes (approved):

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
```diff
@@ -131,6 +131,13 @@ public function respondAction(Request $request)
 
         // Reset state
         $this->getSecondFactorRespondService()->resetRespondState($responseContext);
+
+        // Handle SAML response
+        $httpResponse =  $responseRendering->renderResponse($responseContext, $response, $request);
+
+        $ssoCookieService = $this->get('gateway.service.sso_2fa_cookie');
+        $ssoCookieService->handleSsoOn2faCookieStorage($responseContext, $request, $httpResponse, 'sfo');
```
MKodde, line 139: The writing of the SSO cookie is moved to a point before the ADFS response is sent. The regular response is created at an earlier stage than before, this to be able to create the SSO cookie (it needs the http response object to write the cookie)

This ensures both ADFS and regular SFO authentications can write the sso cookie (when allowed)

src/Surfnet/StepupGateway/GatewayBundle/Service/GlobalViewParameters.php
```diff
@@ -55,8 +55,12 @@ public function __construct(TranslatorInterface $translator, array $locales, arr
...
      */
-    public function getSupportUrl()
+    public function getSupportUrl(): string
     {
-        return $this->supportUrl[$this->translator->getLocale()];
+        $locale = $this->translator->getLocale();
+        if (array_key_exists($locale, $this->supportUrl)) {
+            return $this->supportUrl[$locale];
+        }
+        return '';
```
MKodde, line 64: The application no longer breaks on a missing language in the `stepup_language` cookie.

pablothedude (approved):

PR #296, opened 2023-09-06
Require step up auth when cookie is broken

MKodde:
When the cookie can not be decrypted. Or is otherwise broken, missing, or has gone fishing. The requirement was that the end user is asked for a new step up authentication. Previously this would only be the case when the cookie could not be found. Now the other scenarios when a cookie can't be read is are also deferred to that behavior. Previously a 500 error would be served to the user.

See: https://www.pivotaltracker.com/story/show/183402734/comments/238279428


phavekes (approved):
:shipit:

PR #301, opened 2023-09-11
Reset the SSO cookie on every real 2FA authentication

MKodde:
When ForceAuthn is used, the existing SSO cookie must be overwritten (if SP permits this) This was not yet done but was stated as a requirement in:

`https://www.pivotaltracker.com/story/show/183402542`

> `Each eligible token authentication sets the SSO cookie, thereby overwriting a previous value of the SSO cookie.`

To achieve this goal, we need to know for every authentication if it was a cookie verified authentication, or an actual Second Factor token authentication. When the cookie is used, the state handler would be marked with this fact. And that flag is now used to determine if we need to overwrite the cookie. For example when force authn is used.

The ticket can be found here: `https://www.pivotaltracker.com/story/show/186011497`

As a 'bonus' a remaining chore was also addressed in this PR, dealing with 
[Prevent double log message](https://github.com/OpenConext/Stepup-Gateway/pull/301/commits/0d330e3b498c70d0a684144cf2012c303bfaa88c)

As stated in: `https://www.pivotaltracker.com/story/show/186011501`


src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
```diff
@@ -265,7 +265,7 @@ public function isSecondFactorVerified()
         return $this->get('selected_second_factor_verified') === true;
     }
 
-    public function setVerifiedBySsoOn2faCookie(bool $isVerifiedByCookie): self
+    public function setVerifiedBySsoOn2faCookie(?bool $isVerifiedByCookie): self
```
pablothedude, line 268: Using an unset method would make things clearer then using a nullable parameter.

pablothedude (approved):

src/Surfnet/StepupGateway/GatewayBundle/Saml/Proxy/ProxyStateHandler.php
MKodde: You are absolutely right, changing that :+1:

PR #303, opened 2023-09-18
Remove "Stepup Gateway" title from header

phavekes:
Users don't know what this is, and don't need this information.

Old:
image

New:
image
image
image


pmeulen:
Makes sense. The other option would be to make the "Stepup Gateway" text configurable, but then the questions is what to put there? The logo already provides this information.

pmeulen (approved):

PR #310, opened 2023-11-16
GHA: Add action to build and publish a test container

quartje:
This action will create a test image and push it to the GH container registry. It uses a GH action created by ourselves. See: https://github.com/OpenConext/build-and-publish-test-container

We should think of a better name of the published package. I suggest we use the same tagging strategy as for prod images, and add _test to the container name:
ghcr.io/openconext/stepup-gateway/stepup-gateway_test:4.5.6
ghcr.io/openconext/stepup-gateway/stepup-gateway_test:latest


config/packages/smoketest/doctrine.yaml
```diff
@@ -3,10 +3,7 @@ doctrine:
...
-        port:     "%database_port%"
-        host:     db.stepup.example.com
-        dbname:   gateway
-        user:     gateway
-        password: gateway
-        charset:  UTF8
+        dbname: gateway_test
+        host: mariadb
+        password: gw_deploy_secret
+        user: gw_deploy_user
```
MKodde, line 9: @quartje I think you broke the database tests with this change (at least that's my hunch). Are you able to look at this, or should @parijke or me look into this?

MKodde (changes requested):

MKodde:
Sorry for passing the blame to you @quartje 

Turned out that the ci tests (`test-integration`) stil where running on the `smoketest` env. And your changes, required to get the global behat tests running, broke the GW specific behat tests. 

Those GW specific behat tests now run in the `test` env. And the global tests can still run on `smoketest`. Makes sense.

MKodde (approved):
@parijke Given my eventual involvement in this PR. Can you also do a review?

PR #311, opened 2023-12-18
Fix missing negation in test for issuing "failed" log message

thijskh:
The code checks if the response `isSuccess` but then changes the log message to that the response is failed. This is likely a missing negation in the test.

Note: this is from reading the code, I did not perform actual tests on the before and after situation.


MKodde (approved):
Thanks for pinching this! :crab:

MKodde:
Rebased this to my intermediate 4.2 hotfix branch. After merging and releasing there, I will return this change back to the main branch.

Issue #312, opened 2023-12-20
No javascript not included in SAML Response to SFO authentication requests from ADFS Plugin 


pmeulen:
The response is rendered twice for ADFS, and only the second response is used:

1. https://github.com/OpenConext/Stepup-Gateway/blob/4.2.3/src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php#L136
2. https://github.com/OpenConext/Stepup-Gateway/blob/4.2.3/src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php#L152-L160

The Because Twig only sends javascript once, the second response does not contain javascript. See https://github.com/symfony/webpack-encore-bundle/blob/v1.17.1/src/Asset/EntrypointLookup.php#L84-L87

Introduced in https://github.com/OpenConext/Stepup-Gateway/commit/0f9225a4bdb79f0249898ab8f7a92a43c5a0602c

PR #313, opened 2023-12-20
Ensure consume_assertion.html.twig is rendered once

MKodde:
Previously we would always render the regular consume_assertion.html.twig template, and additionally render the adfs variant when dealing with an ADFS response. That caused the JS from not ending up in the template, as Twig concluded that the (reused) JS was already on the page (as it was just rendered).

This ensures the render method is called only once. Circumventing the problem. But not realy solving the issue at hand. Which is that the controller logic is getting growingly more complex and harder to understand.

For that problem, I'm opening a [technical debt issue in Pivotal](https://www.pivotaltracker.com/story/show/186717176).

See: https://github.com/OpenConext/Stepup-Gateway/issues/312
TD ticket: https://www.pivotaltracker.com/story/show/186717176

### DIFF
--- a/src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
+++ b/src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
@@ -132,24 +132,13 @@ class SecondFactorOnlyController extends Controller
         // Reset state
         $this->getSecondFactorRespondService()->resetRespondState($responseContext);
 
-        // Handle SAML response
-        $httpResponse =  $responseRendering->renderResponse($responseContext, $response, $request);
-
-        $ssoCookieService = $this->get('gateway.service.sso_2fa_cookie');
-        $ssoCookieService->handleSsoOn2faCookieStorage($responseContext, $request, $httpResponse);
-
-        // We can now forget the selected second factor.
-        $responseContext->finalizeAuthentication();
-
-        // Check if ADFS response
+        // Check if ADFS response, if it is, we use the ADFS ACS twig template
         $adfsParameters = $this->getSecondFactorAdfsService()->handleAdfsResponse($logger, $responseContext);
-
         if (!is_null($adfsParameters)) {
             // Handle Adfs response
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/Gateway/AdfsService.php
```diff
@@ -77,15 +77,13 @@ public function handleAdfsRequest(LoggerInterface $logger, Request $httpRequest,
     }
 
     /**
-     * This method detectds if we need to return a ADFS response, If so ADFS parameters are returned.
```
MKodde, line 80: Did some boyscout work in this class while reading through the code path.

:evergreen_tree: :broom:

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
```diff
@@ -132,33 +133,33 @@ public function respondAction(Request $request)
...
-            return $this->render(
+            $httpResponse = $this->render(
                 '@SurfnetStepupGatewaySecondFactorOnly/adfs/consume_assertion.html.twig',
                 [
                     'acu' => $responseContext->getDestinationForAdfs(),
                     'samlResponse' => $xmlResponse,
                     'adfs' => $adfsParameters,
                 ]
             );
+        } else {
```
MKodde, line 150: The else expression ensures the render method is run but once. Another way of fixing this would have been to move this logic to a separate service. Or by decoupling the two code paths even further (dedicated ADFS routes).

See the TD ticket linked in the story to join the discussion in how to fix this better.

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Controller/SecondFactorOnlyController.php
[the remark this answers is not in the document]
MKodde: For point 1, good optimization. My personal preference is to remove the second (obsolete call to load the ResponseRendering object from the container). 

Regarding point 2: I agree. Having additional controller logic will not fix this issue.

PR #315, opened 2024-02-01
Prevent state issues by actually clearing the state

MKodde:
As we no longer use the namespaced attribute bag, we now need to manually clear the gssp prefixed values from the state handler.

This clear method was not implemented correctly, leaving all values in state. Creating some state confusion issues when executing sequential authentication or registrations with a GSSP token of a certain type.

By clearing the state, those issues are resolved

### DIFF
--- a/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Tests/Saml/StateHandlerTest.php
+++ b/src/Surfnet/StepupGateway/SamlStepupProviderBundle/Tests/Saml/StateHandlerTest.php
@@ -1,0 +1,67 @@
+<?php
+
+/**
+ * Copyright 2024 SURFnet bv
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupGateway/SamlStepupProviderBundle/Tests/Saml/StateHandlerTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->attributeBag->set('gssp_provider_name/key2', 'value 2');
+        $this->attributeBag->set('gssp_provider_name/key3', 'value 3');
+        $this->attributeBag->set('tiqr/key1', 'tiqr value 1');
+
+        $this->assertCount(4, $this->attributeBag->all());
+
+        $this->stateHandler->clear();
+
+        $this->assertCount(1, $this->attributeBag->all());
+        $this->assertTrue($this->attributeBag->has('tiqr/key1'));
```
pablothedude, line 65: Please also add an assertion to validate if this is the only one:
`$this->assertCount(1, $this->attributeBag->all());`

pablothedude:
This seems a valid solution. Maybe consider to also use a unique value of the SAML flow as the key of the state to be less dependent on the `clear` method call in terms of defense in depth measure.

pablothedude (approved):

MKodde:
Thanks for the feedback. 

Regarding the test feedback, I feel the current test covers your suggestion. As I already test that the other state persists.

parijke:
If we use this in ALL projects, is hit an opportunity to centralize this logic?

PR #316, opened 2024-02-05
Store SSO cookie only when authn was a success

MKodde:
When the user would cancel the authentication, the response rendering service would still write the SSO cookie. Handing out SSO without an actual successful authentication taken place

This was issue was fixed in the three scenarios where a SSO cookie can be created.

1. On the regular gateway authentication
2. On the second factor only respond action
3. When the response is an ADFS response

https://www.pivotaltracker.com/story/show/186951513

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
@@ -219,8 +219,10 @@ class GatewayController extends Controller
 
         $httpResponse = $this->render($view, $parameters);
 
-        $ssoCookieService = $this->get('gateway.service.sso_2fa_cookie');
-        $ssoCookieService->handleSsoOn2faCookieStorage($responseContext, $request, $httpResponse);
+        if ($response->isSuccess()) {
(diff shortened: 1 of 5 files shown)


src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
```diff
@@ -220,7 +220,9 @@ public function renderSamlResponse(
         $httpResponse = $this->render($view, $parameters);
 
         $ssoCookieService = $this->get('gateway.service.sso_2fa_cookie');
```
pablothedude, line 222: Consider moving this line into the if clause.

pablothedude (approved):

src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
parijke: Consider DI for gateway.service.sso_2fa_cookie

MKodde: Good suggestions guys. @pablothedude I'm following your suggestion.
@parijke your suggestion will be part of the SF upgrade party! Don't you worry

PR #319, opened 2024-03-20
Leverage the BaseContainers images for running quality assurance tests

quartje:
The CI tests still used some old container setup.
This is now migrated to the latest containers from the OpenConext-BaseContainer project.

In the future, it would be nice to use the OpenConext-devconf configuration, to prevent special config for the test environment


MKodde:
Awesome work @quartje. This made things so much easier and quicker! Just the way I like it. I have some remarks, which I think @parijke or myself should address;

1. The sso on 2fa behat test now fails. That should be fixed before merging imho. Alternatively. We can accept this, merge it to the SF6 upgrade branch and fix it there. 
2.  I'd love to have separation between the Acceptance tests and the rest of the tests. For example; in tiqr we run the behat tests in a dedicated GHA. That way the quicker tests can run without having to wait on the behat tests to finish. (quicker feedback loop)
3. Some code got reformatted into something that looks a bit migled. Maybe revert those changes? That one is not high on my priority list

PR #320, opened 2024-03-21
Run the local Gateway Behat tests in env=smoketest

MKodde:
This is in preparation of the CI tests running on devconf containers. What was done: 

- The test environment was detached from the behat test env
- Hostnames are updated to the openconext devconf convention (~stepup.example.org~ dev.openconext.local)
- The test fixutres are now writing to the gateway_test database. Having concurrent gateway and global behat tests might lead to data collisions. But nobody should ever do that right..
- An additional sanity check was added to the behat fixture PDO connection class. It will only open a connection when the application runs in smoketest env. This should prevent accidental dev/prod database wipes when a developer tries to run the test in dev/prod
- Updated the readme with some additional test instructions


public/index.php
```diff
@@ -20,7 +20,7 @@
     Request::setTrustedHosts([$trustedHosts]);
 }
 
-$kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$kernel = new Kernel('smoketest', true);
```
parijke, line 23: @MKodde can't we leave it like that? I want to use the debug bar in smoketest mode. What is the point of hardcoding this?

tests/features/bootstrap/FeatureContext.php
```diff
@@ -64,8 +64,8 @@ public static function setupDatabase(BeforeFeatureScope $scope)
     {
         // Generate test databases
         echo "Preparing test schemas\n";
-        shell_exec("/var/www/bin/console doctrine:schema:drop --env=test --force");
-        shell_exec("/var/www/bin/console doctrine:schema:create --env=test");
+        shell_exec("/var/www/html/bin/console doctrine:schema:drop --env=smoketest --force");
```
parijke, line 67: Why not use the $APP_ENV environment variable

tests/features/bootstrap/ServiceProviderContext.php
```diff
@@ -121,8 +121,7 @@ public function anIdPWithEntityID($entityId)
     private function registerSp($entityId, $sfoEnabled)
     {
         $publicKeyLoader = new KeyLoader();
-        // todo: use from services_test.yml
-        $publicKeyLoader->loadCertificateFile('/var/www/ci/certificates/sp.crt');
+        $publicKeyLoader->loadCertificateFile('/config/ssp/sp.crt');
```
parijke, line 124: In line with the other stepup repo's, this should be `/config/gateway` I think

parijke (commented):
Just some thoughts

PR #323, opened 2024-04-24
Upgrade Stepup-Gateway to Symfony 6.4 with PHP 8.2 support

MKodde:
Upgrade the Gateway to use PHP 8.2 and upgrade Symfony to version 6.4

In this rejuvenation task we diverged slightly from our other work. We had less focus om upgrading to new standards and increasing type safety. Time was a factor in this one. 

I composed this PR from previous work of @parijke and added my own finalization to it. This resulted in a functioning Gateway, with quite a lot of corners cut. For example, expect more remaining deprecation warnings, and a significantly bigger set of PHPStan baseline entries. We aim to address these 'issues' at a later point in time.


MKodde (commented):
Thanks for your final work + review work @parijke 

Lets ship this one to main. And fix any regressions (if any) from there.

PR #328, opened 2024-06-19
Bugfix/fix ci docker

quartje:
The new CI setup failed, since the wrong network name was used in the compose.yml .


parijke (approved):
Looks like its working, ship it

PR #333, opened 2024-07-10
Refrain from rendering a non existing template

MKodde:
The old Twig naming convention was not updated for the ADFS consume assertion twig template.

This should be the last instance where this was not updated in the Gateway

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -102,7 +102,9 @@
             "@phpstan",
             "@phpcs",
             "@phpmd",
-            "@test"
+            "@test",
+            "@behat-adfs",
+            "@behat-metadata"
         ],
         "phplint": "./ci/qa/phplint",
         "validate-lockfile": "./ci/qa/validate",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -102,7 +102,9 @@
             "@phpstan",
             "@phpcs",
             "@phpmd",
-            "@test"
+            "@test",
+            "@behat-adfs",
+            "@behat-metadata"
```
MKodde, line 107: I enabled 2 behat test features that where easily brought back to life. Most important of which are the ADFS tests.

tests/features/bootstrap/FeatureContext.php
```diff
@@ -162,6 +162,9 @@ public function iEnterTheOtp(): void
     public function iEnterTheSmsVerificationCode(): void
     {
         $cookieValue = $this->minkContext->getSession()->getDriver()->getCookie('smoketest-sms-service');
+        if ($cookieValue === null) {
+            throw new RuntimeException('Unable to load the smoketest-sms-service cookie');
+        }
```
MKodde, line 167: Todo, this case is now occurring as the current Behat setup does not use the SMS service test double. Which would set the cookie in the first place.

I created a chore to fix this another time.

https://www.pivotaltracker.com/story/show/187940465

PR #470, opened 2025-11-04
Composer update patch versions

johanib:
## 6.0.0
- Upgrade from Symfony 6.4 to Symfony 7.3
- Maintenance release, Update tooling, packages

Upgrade instructions:
 - Ensure the `database_server_version` setting in parameters.yaml is up-to-date with the production db server version. e.g. `10.6.23-MariaDB`

Attention:
 - `jms/translation-bundle` seems EOL, it relies on `doctrine/annotations`, which is abandoned. See https://github.com/OpenConext/Stepup-Gateway/issues/471


composer.json
```diff
@@ -22,65 +22,67 @@
...
-        "phpunit/phpunit": "^9.6",
-        "ramsey/uuid": "^4",
-        "rector/rector": "^1.0",
-        "sebastian/exporter": "^4.0",
-        "sebastian/phpcpd": "^6.0",
-        "slevomat/coding-standard": "^8.15",
-        "squizlabs/php_codesniffer": "^3.4",
+        "dmore/chrome-mink-driver": "2.9.*",
+        "friends-of-behat/mink-extension": "^2.7.5",
+        "friends-of-behat/symfony-extension": "^2.6.2",
```
MKodde: I see you are configuring the version constraints somewhat stricter than before. I always have to look up the specifics but this is what the Composer docs say about the chevron version pinning:

```
The ^ operator behaves very similarly, but it sticks closer to semantic versioning, and will always allow non-breaking updates. For example ^1.2.3 is equivalent to >=1.2.3 <2.0.0 as none of the releases until 2.0 should break backwards compatibility. For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0 and ^0.0.3 as >=0.0.3 <0.0.4.
```

[source](https://getcomposer.org/doc/articles/versions.md#caret-version-range-)

So where we previously allowed version updates al the way to the next major release, you now constrain it up till the next minor.

Stricter might be better, but also possibly require more frequent attention during updates?

⚖️

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 5.2.0
+- Upgrade from Symfony 6.4 to Symfony 7.3
+- Maintenance release, Update tooling, packages
```
MKodde: If possible, convention asks for a PR reference to each feature/maintenance task in each release.

Also, speculating, should this not be a major release as we are upgrading to a new major version of Symfony? Or maybe no significant platform requirements changed?  In that case this can remain a minor 👍 

For example; the user of this version needs to update some config; which can be considered a breaking change unless the config update is non breaking.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+
+Upgrade instructions:
+ - Ensure the `database_server_version` setting in parameters.yaml is up-to-date with the production db server version. e.g. `10.6.23-MariaDB`
```
MKodde: 👍

composer.json
johanib: Since gateway is not a library, I used composer bump, which handles the version constraints for me. Afaik, this generally works well. Dependabot will still bump the versions anyway.

We can discuss

composer.json
```diff
@@ -16,72 +16,73 @@
...
-        "surfnet/stepup-bundle": "^6.0",
-        "surfnet/yubikey-api-client-bundle": "^5.0",
-        "symfony/config": "6.4.*",
-        "symfony/console": "6.4.*",
-        "symfony/expression-language": "6.4.*",
-        "symfony/flex": "^2",
-        "symfony/form": "6.4.*",
-        "symfony/http-foundation": "^6",
+        "surfnet/stepup-bundle": "dev-feature/maintenance_nov_25",
+        "surfnet/yubikey-api-client-bundle": "dev-feature/maintenance-nov-25",
```
johanib: Merge libraries first, than include proper release before merging.

PR #475, opened 2025-11-19
Update npm packages & tooling

johanib:
This PR updates the npm packages to the latest majors.
It removes jest, as it was unused.
It upgrades bootstrap from 3 > 5.
ESlint is upgraded to version 9.
And others..

Some deliberation took place on legacy browser support.
Currently, gateway uses almost no javascript, except for jquery. Before this change, jquery 3 was already used. Jquery 3 is not supported by old browsers at all. So I do not think, js compatibility should be affected. No new javascript code was added.
As for css, autoprefix was added to ensure old browser fallbacks as needed, in response to the bootstrap 5 upgrade.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 6.0.0
 - Upgrade from Symfony 6.4 to Symfony 7.3
 - Maintenance release, Update tooling, packages
+- Update npm packages and frontend tooling.
 
 Upgrade instructions:
  - Ensure the `database_server_version` setting in parameters.yaml is up-to-date with the production db server version. e.g. `10.6.23-MariaDB`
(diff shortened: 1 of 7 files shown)


MKodde (approved):
I have little comments about the code changes. Awesome to see you upgrade these packages and seeing them actually work well! Chapeau!

Questions after clicking through the /preview pages;

1. I could not view the previews in env = `test`. So either fix that, or only allow in `smoketest`. Either one is fine with me. My personal preference would be for them to also be available in `dev`.
2. Could you drop the popper js dependency? Or is it actually used by Gateway? I could not find it, Nor remember where/when we used it 🤔 
3. Suggestion: add all available SF types to the WAYG screen and maybe fix missing icons?

MKodde (approved):

PR #481, opened 2025-11-25
Improve browser tab handling in MinkContext

johanib:
Prior to this change, some behat tests would randomly fail. Upon investigationg using screenshots, a wrong page was loaded into the tab.

This change improves tab handling by handling:

* Handles slow tab creation by using a wait
* Handles out of order creation better


johanib:
Before this change, the the 'problem' tests would fail ~10-20% of the time. After the change, running the problem tests 200 times caused 1 failure.

PR #483, opened 2025-11-26
Restore bootstrap 3 styling

johanib:
Prior to this change, the templates used the bootstrap 5 styling.

This change reverts the used elements (buttons, links and font) to represent bootstrap 3 styling.

Also adjusted some margins to match the original for consistency.

### DIFF
--- a/public/scss/application.scss
+++ b/public/scss/application.scss
@@ -7,6 +7,11 @@
 @import "../../node_modules/bootstrap/scss/bootstrap";
 
 body {
+  // Start Bootstrap 3-like font overrides
+  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-size: 15px;
+  // End Bootstrap 3-like font overrides
+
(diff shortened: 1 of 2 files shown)


MKodde (approved):
I approve of this, as long as this is a 'best effort' task. If it should be pixel perfect I'd need to do some more investigating. 

Are you aware the floating yubikey submit button is still lingering just outside the viewport on the Yubi OTP screen?

Image

johanib:
@pmeulen See the screenshot above. 
On the YubiKey code screen, there is a scrollbar on the bottom. If you scroll to the right, a submit button will become visible. 

I'm not sure why this button is moved off screen instead of being hidden. Maybe for accessibility reasons. 

I think the bug is that the scroll bar is visible. 
Should we create a bug ticket for this issue?
This is not a regression in the update, the current version also has the scroll bar.

See https://github.com/OpenConext/Stepup-Gateway/blame/main/public/scss/application.scss#L71

PR #493, opened 2025-12-03
Upgrade Symfony 7.3 > 7.4

johanib:
Prior to this change, Symfony was on version 7.3, which is not an LTS release. This makes it more difficult to quickly rollout security hotfixes.

This change takes SF to 7.4 LTS and resolves the deprecations introduced in 7.4.

### DIFF
--- a/src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
+++ b/src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
@@ -497,6 +497,10 @@ class SecondFactorController extends ContainerController
         Request $request,
     ): Response|RedirectResponse {
         $authenticationMode = $request->query->getString('authenticationMode', '');
+        if ($authenticationMode === '') {
+            $authenticationMode = $request->request->get('authenticationMode');
+        }
+
         $this->supportsAuthenticationMode($authenticationMode);
         $context = $this->getResponseContext($authenticationMode);
         $originalRequestId = $context->getInResponseTo();


src/Surfnet/StepupGateway/GatewayBundle/Controller/GatewayController.php
```diff
@@ -174,10 +170,19 @@ public function sendAuthenticationCancelledByUser(): Response
...
-        if (!$request->get('authenticationMode', false)) {
+
+        if ($request === null) {
+            throw new RuntimeException('No request');
+        }
+
+        $authenticationMode = $request->query->getString('authenticationMode');
+        if ($authenticationMode === '') {
+            $authenticationMode = $request->request->getString('authenticationMode');
+        }
```
MKodde: I dont think we'll need the fallback to the post request params. AFAIR the authentication mode is always passed as a query parameter.

src/Surfnet/StepupGateway/GatewayBundle/Controller/SecondFactorController.php
```diff
@@ -498,10 +496,14 @@ public function verifyYubiKeySecondFactor(Request $request): Response
     public function verifySmsSecondFactor(
         Request $request,
     ): Response|RedirectResponse {
-        if (!$request->get('authenticationMode', false)) {
+        $authenticationMode = $request->query->get('authenticationMode');
+        if ($authenticationMode === '') {
+            $authenticationMode = $request->request->get('authenticationMode');
+        }
```
MKodde: Fallback to post params should not be required here, also goes for the other situations.

PR #559, opened 2026-03-05
rename yubikey entry form to solve #558

### LINKED ISSUE #558: When using the Apple password manager, yubikey entry is broken
When installing the [icloud password manager](https://chromewebstore.google.com/detail/icloud-wachtwoorden/pejdijmoenmkgeppbflobdenhhabjlaj) in Chrome or Vivaldi on mac, the yubikey otp entry fails. It looks to the user the [enter] isn't parsed. Manually adding a enter does not work either. 

The browser console shows: 

Image

Disabling the extension fixes the issue.

Adding `autocomplete="off"` to the input also seems to fix it

phavekes:


kayjoosten (dismissed):
In overleg een patch gestuurd met een oplossing voor het probleem

src/Surfnet/StepupGateway/GatewayBundle/Form/Type/VerifyYubikeyOtpType.php
```diff
@@ -28,12 +28,11 @@ class VerifyYubikeyOtpType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
-        $builder->add('otp', TextType::class, [
+        $builder->add('yubikeyInput', TextType::class, [
             'label' => /** @Ignore */ false,
             'required' => true,
             'attr' => [
                 'autofocus'    => true,
-                'autocomplete' => 'off',
```
johanib, line 36: This autocomplete `off` has been here for 11 years. Is the removal intentional?

Probably should not be removed.

public/scss/authentication.scss
```diff
@@ -20,7 +20,7 @@ body.authentication {
     }
   }
 
-  form[name="gateway_verify_yubikey_otp"] {
+  form[name="gateway_verify_yubikey"] {
```
johanib, line 23: Should probably also update the tag of `gateway.form.verify_yubikey:` in `src/Surfnet/StepupGateway/GatewayBundle/Resources/config/services.yml`

johanib (changes requested):
Please cleanup commits before merging (or write proper commit description during squah merge)

johanib:
I tested this manually, and it does fix the issue :confetti_ball: 

We should probably also fix this in RA and SelfService, as the issue occurs in those projects.

src/Surfnet/StepupGateway/GatewayBundle/Form/Type/VerifyYubikeyOtpType.php
kayjoosten: This one is intentional, the autocomplete is something you can set but will be ignored by password managers. So its kinda dead. if you rather have me but it back i will.

johanib: If its intentional, I'm fine with removing it :+1:

johanib (approved):

PR #624, opened 2026-06-29
feat: read mdui:UIInfo from AuthnRequest and forward service name to GSSP

kayjoosten:
## Summary

Closes #591.

When EngineBlock sends a `mdui:UIInfo` extension in the AuthnRequest, Gateway now reads the `mdui:DisplayName` values and forwards the service name in a `mdui:UIInfo` extension on the proxy AuthnRequest sent to the GSSP.

- Feature-flagged via `enable_service_name_from_saml_authnrequest` (disabled by default, configure in `parameters.yaml`)
- Applies to both SSO (`GatewayBundle`) and SFO (`SecondFactorOnlyBundle`) flows
- Priority: middleware `service_name` (from SP configuration) overrides `mdui:DisplayName` from the AuthnRequest

**Priority table**

| Middleware `service_name` | AuthnRequest `mdui:DisplayName` | Result |
|---|---|---|
| No | Yes | AuthnRequest value forwarded |
| Yes | Yes | Middleware value used |
| No | No | No extension added |
| Yes | No | Middleware value forwarded |

## Changes

- `UiInfoExtensionHelper` — parses and builds `mdui:UIInfo` SAML extensions
- `DisplayName` — value object for type-safe lang/value pairs (serializes as plain arrays in session)
- `FeatureConfiguration` — named accessor replacing raw `bool` constructor args
- `ProxyStateHandler` — stores/retrieves `DisplayName[]` in session state
- `ResponseContext::resolveServiceDisplayNames()` — applies priority logic with real SP data
- SSO + SFO `LoginService` — parses UIInfo from incoming request via `UiInfoExtensionHelper::parseAndStore`
- `SecondFactorVerificationService` — forwards resolved display names to GSSP proxy AuthnRequest
- `parameters.yaml.dist` — documents the new flag

## Test plan

- [x] `composer test` — 258 tests pass
- [x] `composer phpstan` — clean
- [x] `composer phpcs` — clean
- [x] `composer phpmd` — clean
- [x] `composer phplint` — clean
- [x] New tests: `ProxyStateHandlerDisplayNameTest`, `ResponseContextResolveDisplayNamesTest`, extended `UiInfoExtensionHelperTest`, SFO branch in `SecondFactorVerificationServiceTest`
- [x] Behat: requires Docker environment (`devconf/stepup`)
- [x] Enable flag in `parameters.yaml` and verify service name appears in GSSP AuthnRequest with EngineBlock sending `feature_stepup_send_service_name=true`


src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
```diff
@@ -369,4 +370,35 @@ public function getRequestServiceProvider(): ?string
...
+     *
+     * @return DisplayName[]
+     */
+    public function resolveServiceDisplayNames(bool $includeAuthnRequestFallback = true): array
+    {
+        $sp = $this->getServiceProvider();
+        if ($sp !== null) {
+            $serviceName = $sp->getServiceName();
+            if ($serviceName !== null && $serviceName !== '') {
+                return [new DisplayName('en', $serviceName)];
```
johanib: **Middleware `service_name` is hardcoded to `lang="en"`**

The middleware `service_name` is a single string with no language metadata, but it's emitted as `new DisplayName('en', $serviceName)`. For non-English deployments the GSSP will receive an inaccurate `xml:lang`, which may affect how the GSSP picks/labels the name.

Suggested approach: If a locale is available (e.g. preferred locale in state) use it; otherwise document that `en` is a deliberate neutral default.

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/SecondFactorVerificationService.php
```diff
@@ -92,6 +93,18 @@ public function sendSecondFactorVerificationAuthnRequest(
         );
         $authnRequest->setSubject($subjectNameId);
 
+        $activeResponseContext = $responseContextServiceId === 'second_factor_only.response_context'
```
johanib: Probably should introduce a constant for `second_factor_only.response_context`

src/Surfnet/StepupGateway/GatewayBundle/Saml/DisplayName.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function __construct(
+        public readonly string $lang,
+        public readonly string $value
+    ) {
+    }
+
+    public function toArray(): array
+    {
+        return ['lang' => $this->lang, 'value' => $this->value];
```
johanib, line 38: Why is the locale here? I thought EB would set the displayname based on the current locale?

src/Surfnet/StepupGateway/GatewayBundle/Saml/ResponseContext.php
johanib: Possible to fallback to default locale instead of hardcoded 'en' locale?

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
```diff
@@ -0,0 +1,114 @@
...
+            $displayNames[] = new DisplayName($lang, $value);
+            if (count($displayNames) >= self::MAX_DISPLAY_NAMES) {
+                break;
+            }
+        }
+
+        return $displayNames;
+    }
+
+    private static function isAcceptableDisplayName(string $lang, string $value): bool
```
johanib: Lang? (see other comment in `DisplayName`)

Edit: Also multiple displaynames? 
I'm confused. Is this part of the chain compatible with multiple locales?

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
```diff
@@ -0,0 +1,114 @@
...
+use Surfnet\StepupGateway\GatewayBundle\Saml\Proxy\ProxyStateHandler;
+
+class UiInfoExtensionHelper
+{
+    private const MDUI_NAMESPACE = 'urn:oasis:names:tc:SAML:metadata:ui';
+    private const MDUI_PREFIX = 'mdui';
+
+    private const MAX_DISPLAY_NAMES = 10;
+    private const MAX_LANG_LENGTH = 35;
+    private const MAX_VALUE_LENGTH = 1024;
```
johanib: Lets discuss:

Additional spec gaps the PR misses (service-name processing / restrictions)                                                                                           
                                                                                                                                                                      
The RFC mandates sanitization the PR doesn't do. PR uses  MAX_VALUE_LENGTH=1024 ,  MAX_LANG_LENGTH=35 ,  MAX_DISPLAY_NAMES=10  — none match the spec, which requires: 
                                                                                                                                                                      
•  MAX_CHARACTERS = 40 ; truncate to 39 + append ellipsis  …  (U+2026)                                                                                                
• trim leading/trailing whitespace                                                                                                                                    
• collapse consecutive whitespace to a single space                                                                                                                   
• strip control/format characters                                                                                                                                     
                                                                                                                                                                      
None of this is present. (Motivation: the name goes into SMS/Tiqr push, 160/70-char limits.)

johanib (commented):
I lost the overview of what we discussed in this feature. So this review is a bit messy. I think the main question is:

Where can I find the main, definitive spec we agreed on for the service name feature?

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
johanib: discussed: This is available in one of the bundels?

src/Surfnet/StepupGateway/GatewayBundle/Saml/DisplayName.php
kayjoosten: This is the middleware `service_name` override, separate from the EB/mdui path. Middleware's name is a single untranslated string with no locale of its own. We tag it with Gateway's own session locale instead, normalized to `en`/`nl` since the session stores `en_GB`/`nl_NL`. Fixed in 462b341e.

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
kayjoosten: Partially. gssp-bundle has ServiceNameFormatter/ServiceNameResolver. gssp-example and tiqr reuse it. Gateway can't depend on gssp-bundle, wrong direction. Its own sanitizeValue() here is intentional. Found 2 real bugs checking this: gssp-bundle's formatter truncated to 41 chars not 40, off by one against the RFC. Webauthn had its own separate duplicate with the same bug instead of reusing gssp-bundle's. Fixed both, pushed.

kayjoosten:
OpenConext/Stepup-Gateway#587. Now linked in code next to MAX_CHARACTERS. MAX_VALUE_LENGTH is SAML2's own bound, not this RFC. MAX_LANG_LENGTH and MAX_DISPLAY_NAMES aren't spec derived at all, just internal limits. Also commented now.

config/openconext/parameters.yaml.dist
```diff
@@ -189,4 +189,8 @@ parameters:
     fallback_gssp_subject_attribute: 'urn:mace:dir:attribute-def:mail'
 
     # The user attribute to use to determine the user's home institution
-    fallback_gssp_institution_attribute: 'urn:mace:terena.org:attribute-def:schacHomeOrganization'
\ No newline at end of file
+    fallback_gssp_institution_attribute: 'urn:mace:terena.org:attribute-def:schacHomeOrganization'
+
+    # When enabled, Gateway reads mdui:DisplayName from incoming AuthnRequest extensions and forwards
+    # it in the proxy AuthnRequest to the GSSP. Middleware service_name takes priority if set.
+    enable_service_name_from_saml_authnrequest: false
```
johanib: `append_service_name_to_authnrequest`

rfcs/RFC-Servicename-from-engie-to-Tiqr.md
johanib: rm

src/Surfnet/StepupGateway/GatewayBundle/Service/Gateway/LoginService.php
```diff
@@ -66,7 +68,8 @@ public function __construct(
         LoaResolutionService $loaResolutionService,
         ServiceProvider $hostedServiceProvider,
         IdentityProvider $remoteIdp,
-        RedirectBinding $redirectBinding
+        RedirectBinding $redirectBinding,
+        private readonly FeatureConfiguration $featureConfiguration = new FeatureConfiguration()
```
johanib: `private readonly FeatureConfiguration $featureConfiguration` > Inject required symfony service from service container, no default init here

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/LoginService.php
```diff
@@ -31,6 +35,11 @@ class LoginService
 {
     private const RESPONSE_CONTEXT_SERVICE_ID = 'gateway.proxy.response_context';
 
+    // Same value as \Surfnet\SamlBundle\SAML2\AuthnRequest::PARAMETER_RELAY_STATE. Kept as a
+    // local constant (rather than referencing that class) to avoid growing this class's
+    // CouplingBetweenObjects count for a single HTTP query parameter name.
+    private const PARAMETER_RELAY_STATE = 'RelayState';
```
johanib: just use `\Surfnet\SamlBundle\SAML2\AuthnRequest::PARAMETER_RELAY_STATE`

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/LoginService.php
```diff
@@ -137,4 +165,35 @@ public function singleSignOn(Provider $provider, Request $httpRequest)
...
     }
+
+    /**
+     * Resolve display names applying the same priority as
+     * ResponseContext::resolveServiceDisplayNames(): middleware service_name
+     * overrides any mdui:DisplayName already present on the original request.
+     *
+     * @return DisplayName[]
+     */
+    private function resolveServiceDisplayNames(
```
johanib: Based on what we discussed earlier:

```
Here's how I'd restructure it. Three concerns are currently tangled in one static grab-bag class: wire-format mapping, 
which-name-wins policy, and session storage.                                                                           
                                                                                                                       
1. Extract the policy into one injectable service                                                                      
                                                                                                                       
final class ServiceDisplayNameResolver                                                                                 
{                                                                                                                      
    public function __construct(                                                                                       
        private FeatureConfiguration $featureConfiguration,                                                            
        private SamlEntityService $samlEntityService,                                                                  
    ) {}                                                                                                               
                                                                                                                       
    /** @param DisplayName[] $fromRequest  @return DisplayName[] */                                                    
    public function resolve(?string $spEntityId, array $fromRequest): array                                            
    {                                                                                                                  
        if (!$this->featureConfiguration->isServiceNameFromSamlAuthnRequestEnabled()) {                                
            return [];                                                                                                 
        }                                                                                                              
        if ($spEntityId !== null && $this->samlEntityService->hasServiceProvider($spEntityId)) {                       
            $names = $this->samlEntityService->getServiceProvider($spEntityId)->getServiceNames();                     
            if ($names !== []) {                                                                                       
                return DisplayName::fromLocaleMap($names);                                                             
            }                                                                                                          
        }                                                                                                              
        return $fromRequest;                                                                                           
    }                                                                                                                  
}                                                                                                                      
                                                                                                                       
Both flows call this; only the source of the two inputs differs (session state vs. live request). The flag lives inside
 — no  bool  parameter threaded through  ResponseContext , no call-site  if . Unknown-SP handling becomes uniform.     
 ResponseContext::resolveServiceDisplayNames()  becomes a one-line delegation (or disappears).                         
                                                                                                                       
2. Make the extension helper a real mapper, and instance-based                                                         
                                                                                                                       
 UiInfoExtensionMapper  with  read(Extensions): DisplayName[]  and  applyTo(Extensions, DisplayName[]): Extensions  —  
where an empty array means strip. Then every call site is one line:                                                    
                                                                                                                       
$proxyRequest->setExtensions(                                                                                          
    $this->uiInfo->applyTo($originalRequest->getExtensions(), $displayNames)                                           
);                                                                                                                     
                                                                                                                       
Flag-off ⇒ resolver returns  []  ⇒ UIInfo stripped, by construction rather than by a parallel  else  branch.           
 merge / strip / build  collapse into one contract.                                                                    
                                                                                                                       
3. Drop  parseAndStore()                                                                                               
                                                                                                                       
It makes the SAML mapper depend on  ProxyStateHandler  (session) — a layering inversion. Let the caller do it:         
 $stateHandler->setDisplayNamesFromRequest(...$this->uiInfo->read($request->getExtensions())) .                        
                                                                                                                       
4. Push locale mapping into the domain                                                                                 
                                                                                                                       
 DisplayName::fromLocaleMap(array $serviceNames)  (or have  ServiceProvider::getServiceNames()  return  DisplayName[] )
 removes  buildDisplayNamesFromLocaleMap  and  normalizeToPrimarySubtag  from the helper's public surface.             
                                                                                                                       
5. Keep sanitization at the outbound boundary                                                                          
                                                                                                                       
The 40-char/ellipsis rule is presentation for SMS/push, so it belongs in the writer side of the mapper — that part of  

```

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/Gateway/LoginService.php
```diff
@@ -132,6 +135,10 @@ public function singleSignOn(Request $httpRequest, ReceivedAuthnRequest $origina
 
         $this->stateHandler->markAuthenticationModeForRequest($originalRequestId, 'sfo');
 
+        if ($this->featureConfiguration->isServiceNameFromSamlAuthnRequestEnabled()) {
```
johanib, line 139: Johan: Correct call location?
No response object? 
Which flow is this? This reads mdui from request? But is that correct?

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/LoginService.php
```diff
@@ -112,10 +133,17 @@ public function singleSignOn(Provider $provider, Request $httpRequest)
             $provider->getRemoteIdentityProvider()
         );
 
-        if ($originalRequest->getExtensions()) {
-            $proxyRequest->setExtensions($originalRequest->getExtensions());
+        $extensions = $originalRequest->getExtensions();
+
+        if ($this->featureConfiguration->isServiceNameFromSamlAuthnRequestEnabled()) {
+            $displayNames = $this->resolveServiceDisplayNames($originalRequest);
```
johanib: Question: In SFO only context (this?), should the originalRequest be consulted?

johanib (commented):
It would be very valuable to have these mdui scenarios, happy + non happy covered in behat features.

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/SecondFactorVerificationService.php
johanib: Nice, but still referenced here, \Surfnet\StepupGateway\SecondFactorOnlyBundle\Controller\SecondFactorOnlyController::getResponseContext would be nice to replace as wel. Not blocking

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
```diff
@@ -0,0 +1,235 @@
...
+            if ($child->localName !== 'DisplayName' || $child->namespaceURI !== self::MDUI_NAMESPACE) {
+                continue;
+            }
+            $lang = $child->getAttribute('xml:lang');
+            $value = $child->textContent;
+            if (!self::isAcceptableDisplayName($lang, $value)) {
+                continue;
+            }
+            $displayNames[] = new DisplayName($lang, $value);
+            if (count($displayNames) >= self::MAX_DISPLAY_NAMES) {
```
johanib: Why check max 10? We only have 1, and never more than 1.

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionHelper.php
```diff
@@ -0,0 +1,235 @@
...
+     * element is parsed here, stored, and later re-emitted by buildExtensionsWithUiInfo() when
+     * forwarding to the GSSP. Per the RFC (OpenConext/Stepup-Gateway#587), the source that builds
+     * the AuthnRequest is expected to already resolve to a single locale-matched, sanitized name
+     * before sending it — but this parser stays defensive and doesn't assume that, so it still
+     * accepts (and forwards) more than one language if a source sends them.
+     */
+    public static function parseAndStore(ReceivedAuthnRequest $request, ProxyStateHandler $stateHandler): void
+    {
+        $chunks = $request->getExtensions()->getChunks();
+        $displayNames = isset($chunks['UIInfo'])
```
johanib: This sends out multiple displaynames, we should only send out 1.

src/Surfnet/StepupGateway/GatewayBundle/Entity/SamlEntity.php
```diff
@@ -132,6 +132,7 @@ public function toServiceProvider()
         if (isset($decodedConfiguration['set_sso_cookie_on_2fa'])) {
             $configuration['setSsoCookieOn2fa'] = $decodedConfiguration['set_sso_cookie_on_2fa'];
         }
+        $configuration['serviceNames'] = $decodedConfiguration['service_name'] ?? [];
```
johanib: This takes the list of service names from the saml message, and converts in to an array. This should be a string, because this wrongfully assumes there can be multiple.

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/LoginService.php
johanib: Yes. Discussed with Pieter: When original request contains displayname, keep it. If override config is set on the ServiceProvider, use that instead.

src/Surfnet/StepupGateway/SecondFactorOnlyBundle/Service/Gateway/LoginService.php
```diff
@@ -132,6 +136,12 @@ public function singleSignOn(Request $httpRequest, ReceivedAuthnRequest $origina
...
+            $this->stateHandler->setDisplayNamesFromRequest(
+                ...$this->uiInfoMapper->read($originalRequest->getExtensions())
```
johanib, line 141: https://github.com/OpenConext/Stepup-Gateway/commit/96acbcbe36df4a31a419d261f8ecd09d30a26ef8#diff-f3ebae3a49db3422395360d17911ba0509862b53a7004bc6a2f13616d3013445R10-R158

Decision, fact: It is correct that the SAML request uses an array, not a string. So, even though it only passes one language, it still is an array.

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionMapper.php
```diff
@@ -0,0 +1,177 @@
...
+            $displayNames[] = new DisplayName($lang, $value);
+            if (count($displayNames) >= self::MAX_DISPLAY_NAMES) {
+                break;
+            }
+        }
+
+        return $displayNames;
+    }
+
+    private function isAcceptableDisplayName(string $lang, string $value): bool
```
johanib: `isSamlValid`

src/Surfnet/StepupGateway/SamlStepupProviderBundle/Service/Gateway/SecondFactorVerificationService.php
```diff
@@ -92,6 +98,19 @@ public function sendSecondFactorVerificationAuthnRequest(
         );
         $authnRequest->setSubject($subjectNameId);
 
+        $activeResponseContext = $responseContextServiceId === self::SFO_RESPONSE_CONTEXT_SERVICE_ID
+            ? $this->sfoResponseContext
+            : $this->responseContext;
+        $locale = $this->requestStack->getCurrentRequest()?->getLocale() ?? 'en';
```
johanib, line 104: Does getLocale return en_GB or en?

src/Surfnet/StepupGateway/GatewayBundle/Saml/ServiceDisplayNameResolver.php
```diff
@@ -0,0 +1,112 @@
...
+
+        $normalized = [];
+        foreach ($serviceNames as $configuredLocale => $serviceName) {
+            $normalized[DisplayName::normalizeLocale($configuredLocale)] = $serviceName;
+        }
+
+        if (isset($normalized[$primarySubtag])) {
+            return new DisplayName($primarySubtag, $normalized[$primarySubtag]);
+        }
+        if (isset($normalized['en'])) {
```
johanib: isset($normalized['en'] instead of default language? (This is according to RFC?)

src/Surfnet/StepupGateway/GatewayBundle/Saml/DisplayName.php
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupGateway\GatewayBundle\Saml;
+
+class DisplayName
+{
+    public function __construct(
+        public readonly string $lang,
```
johanib, line 26: A value object should not be able to be created with invalid values, so it can never be in an invalid state. 

I'd argue that a empty language, and probably, and empty name, is an invalid value. 
So, throw exception.

src/Surfnet/StepupGateway/GatewayBundle/Saml/ServiceDisplayNameResolver.php
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        if ($spEntityId !== null && $this->samlEntityService->hasServiceProvider($spEntityId)) {
+            $serviceNames = $this->samlEntityService->getServiceProvider($spEntityId)->getServiceNames();
+            if (!empty($serviceNames)) {
+                return $this->selectFromLocaleMap($serviceNames, $locale);
+            }
+        }
+
+        return $this->selectFromDisplayNames($fromRequest, $locale);
```
johanib: Both selectFromLocaleMap and selectFromDisplayNames should have the same functional requirements.

* get for locale
* If not availeble, return en

tests/features/service-name.feature
```diff
@@ -0,0 +1,116 @@
...
+    And an IdP with EntityID https://ssp.dev.openconext.local/saml2/idp/metadata.php
+    And a whitelisted institution dev.openconext.local
+    And a user from "dev.openconext.local" identified by "urn:collab:person:dev.openconext.local:hafthor_bjornsson" with a vetted "tiqr" token
+    When urn:collab:person:dev.openconext.local:hafthor_bjornsson starts an SFO authentication requiring LoA self-asserted with mdui DisplayNames:
+      | locale | name         |
+      | en     | Spoofed Name |
+    Then the AuthnRequest sent to the GSSP should match xpath '//mdui:UIInfo/mdui:DisplayName[@xml:lang="en" and text()="Real Name"]'
+    And the AuthnRequest sent to the GSSP should not match xpath '//mdui:UIInfo/mdui:DisplayName[text()="Spoofed Name"]'
+
+  Scenario: An SFO authentication without any display name sends no mdui:UIInfo at all
```
johanib, line 104: On very first glance, this scenario covers the gssp registration & SF authentication flow. But we have 3 flow? check if we covered all.

Add: If fallback & locale are not available, it returns NO mdui displayname (all 3 flows)
Add: feature flag is disabled, it returns NO mdui displayname (all 3 flows)

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionMapper.php
```diff
@@ -0,0 +1,177 @@
...
+    {
+        return preg_replace('/[\p{Cc}\p{Cf}]/u', '', $lang) ?? $lang;
+    }
+
+    /**
+     * Trims whitespace, collapses consecutive whitespace to a single space, strips control/format
+     * characters, and truncates to MAX_CHARACTERS (with an ellipsis) so the display name fits the
+     * SMS/Tiqr push notifications it's forwarded into.
+     */
+    private function sanitizeValue(string $value): string
```
johanib: Why re-implement when we've got src/Service/ServiceName/ServiceNameFormatter.php in https://github.com/OpenConext/Stepup-gssp-bundle/pull/49/changes?

kayjoosten: the Stepup-gssp-bundle is not in Gateway atm. I dont think its worth adding for 1 sanitizer. But you are right its duplicate code. 

What we could do is move this logic to the saml-bundle that is in the gateway. Or leave it as is. What do you think?

tests/features/service-name.feature
johanib: Covered by SFO :+1

src/Surfnet/StepupGateway/GatewayBundle/Entity/SamlEntity.php
```diff
@@ -132,6 +132,12 @@ public function toServiceProvider()
         if (isset($decodedConfiguration['set_sso_cookie_on_2fa'])) {
             $configuration['setSsoCookieOn2fa'] = $decodedConfiguration['set_sso_cookie_on_2fa'];
         }
+        // service_name is expected to be a locale => name map. Middleware's own validator
+        // (Stepup-Middleware#606) currently only enforces "nullable string", so a plain
```
johanib, line 136: Please cleanup comments

src/Surfnet/StepupGateway/GatewayBundle/Saml/UiInfoExtensionMapper.php
```diff
@@ -0,0 +1,161 @@
...
+ */
+class UiInfoExtensionMapper
+{
+    private const MDUI_NAMESPACE = 'urn:oasis:names:tc:SAML:metadata:ui';
+    private const MDUI_PREFIX = 'mdui';
+
+    // Defensive parsing bounds, not derived from any spec — cap how much an incoming
+    // AuthnRequest's UIInfo extension can make this parser do.
+    private const MAX_DISPLAY_NAMES = 10;
+    private const MAX_LANG_LENGTH = 35;
```
johanib, line 35: Check if can be replaced by by stepup-saml-bundle shared logic

src/Surfnet/StepupGateway/GatewayBundle/Saml/ServiceDisplayNameResolver.php
johanib: there is no default locale currently in Gateway

PR #636, opened 2026-07-14
Fix feature flag incorrectly suppressing middleware service name override

kayjoosten:
## Problem

With the AuthnRequest-service-name feature flag disabled, the middleware-configured `service_name` override for a service provider was not shown during second factor verification either, even though the flag is only meant to control the AuthnRequest mdui:DisplayName fallback.

## Root cause

`SecondFactorVerificationService::sendSecondFactorVerificationAuthnRequest()` wrapped the entire `resolveServiceDisplayNames()` call (both the middleware-override branch and the AuthnRequest-fallback branch) in the `isServiceNameFromSamlAuthnRequestEnabled()` flag check, instead of only gating the AuthnRequest-derived fallback.

## Fix

`ResponseContext::resolveServiceDisplayNames()` now accepts an `includeAuthnRequestFallback` parameter. The middleware override is always considered; only the AuthnRequest fallback is skipped when the parameter is false. `SecondFactorVerificationService` now always calls `resolveServiceDisplayNames()`, passing the feature flag value as this parameter instead of wrapping the whole call.

## Testing

Added tests confirming the middleware override survives with the flag disabled, and that no fallback occurs without it. Verified against the old code that these tests fail (confirming they reproduce the bug). Full suite (265/265), phpcs and phpmd pass.

### DIFF
--- a/config/openconext/parameters.yaml.dist
+++ b/config/openconext/parameters.yaml.dist
@@ -190,3 +190,7 @@ parameters:
 
     # The user attribute to use to determine the user's home institution
     fallback_gssp_institution_attribute: 'urn:mace:terena.org:attribute-def:schacHomeOrganization'
+
+    # When enabled, Gateway reads mdui:DisplayName from incoming AuthnRequest extensions and forwards
+    # it in the proxy AuthnRequest to the GSSP. Middleware service_name takes priority if set.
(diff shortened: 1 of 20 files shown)


kayjoosten:
Closing - fix has been pushed directly to feature/issue-591-service-name-saml-extension instead (see PR #624).

